PR #167, opened 2026-04-22
Shell: fix frozen rotation/drag on Intel iGPU

leaiss:
## Summary

- Shell rotation and drag were frozen on Intel iGPU because the compositor only repainted when an OpenXR app submitted a frame — at 5–10fps on slow iGPU hardware
- Start render timer thread when first 3D IPC client connects (previously only started for 2D capture sessions)
- Remove skip condition that assumed IPC clients would drive rendering fast enough via `layer_commit`
- Guard thread stop so it stays alive while any clients (IPC or capture) are connected
- Replace `Sleep(8)` polling with `WaitForSingleObject(render_wakeup_event, 14)`: precise 14ms interval, instant shutdown, foundation for future drag-responsive signaling
- Fix latent bug: timer thread now updates `last_shell_render_ns` so `layer_commit`'s 14ms throttle correctly accounts for timer-driven renders

Shell UI now renders at ~70fps independent of app framerate. Existing 14ms throttle in `layer_commit` prevents double-renders when both paths fire.

## Test plan

- [ ] Run shell with 2 test apps on Intel iGPU (Windows Graphics Settings → Power saving for `displayxr-service.exe`)
- [ ] Verify window rotation (right-click drag on title bar) is smooth
- [ ] Verify window drag (left-click drag on title bar) is smooth
- [ ] Verify no regression on discrete GPU — shell still renders correctly at full framerate
- [ ] Verify graceful shutdown — service exits cleanly without hanging on thread join

### DIFF
--- a/src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp
+++ b/src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp
@@ -815,9 +815,12 @@ struct d3d11_multi_compositor
 
 	//! @name Capture client render timer
 	//! @{
-	std::thread capture_render_thread;           //!< Timer thread for capture-only rendering
+	std::thread capture_render_thread;              //!< Timer thread for shell rendering
 	std::atomic<bool> capture_render_running{false}; //!< Thread run flag
-	uint32_t capture_client_count{0};            //!< Number of active capture-type slots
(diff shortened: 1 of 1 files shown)


leaiss:
Superseded by #168 (cleaner history)

PR #201, opened 2026-05-04
VK clients in workspace mode signal cross-process fence

leaiss:
## Summary

VK handle apps in shell mode (`DISPLAYXR_WORKSPACE_SESSION=1`) rendered black because the IPC client compositor never advanced the per-client `workspace_sync_fence` that the D3D11 service waits on. The service treated every view as stale and skipped the blit (`stale_views >> 0`, `fence_total_waits = 0` in `bench_shell_present.ps1` output).

The D3D11 client implemented the Phase 2 cross-process fence; the VK client did not. This PR mirrors that path on the VK side.

## Changes

- **`comp_vk_client`**: import the service's `workspace_sync_fence` as a VK timeline semaphore at create; host-signal it after `vkQueueWaitIdle` in `submit_fallback`; ship the value via `comp_ipc_client_compositor_set_workspace_sync_fence_value` so the per-view loop can `ID3D11DeviceContext4::Wait` on it.
- **`comp_vk_glue`**: declare `VK_KHR_timeline_semaphore` as a required device extension via `xrGetVulkanDeviceExtensionsKHR`.
- **`oxr_session_gfx_vk`**: force `timeline_semaphore_enabled = true` in IPC mode (the runtime now requires the extension; v1-path apps don't surface the feature flag any other way).
- **`vk_function_loaders`**: fall back to core `vkSignalSemaphore` / `vkWaitSemaphores` / `vkGetSemaphoreCounterValue` when the KHR aliases aren't exposed (NVIDIA on VK 1.2+ devices behaves this way).
- **`u_sandbox`**: add the same Win32 `GetEnvironmentVariableA` fallback the `XRT_FORCE_MODE` check already has, so `SetEnvironmentVariableA`-set `DISPLAYXR_WORKSPACE_SESSION` reaches /MT-CRT VK apps that miss it via `getenv()`.
- **`tests/CMakeLists`**: link `tests_comp_client_vulkan` against `ipc_client` (matches `tests_comp_client_d3d11` — same bridge symbols now referenced).

Companion change in `displayxr-demo-gaussiansplat` enables `features12.timelineSemaphore` on the demo's `VkDevice`. Other VK demos that use `xrGetVulkanDeviceExtensionsKHR` will need the equivalent feature-enable before they can use this path.

## Test plan
- [x] Build runtime: `scripts\build_windows.bat build`
- [x] Build gauss splat demo (with companion PR)
- [x] Launch shell with 4 gauss splats — all four windows render content (was black)
- [x] Verify service log: `stale_views=0`, `waits_queued=210+` per 10 s window per client
- [x] Mixed workload: 2 cubes + 2 gauss splats in shell — both APIs render correctly
- [ ] Cube-only regression check (D3D11 path unaffected by these changes — but please verify)

### DIFF
--- a/src/xrt/auxiliary/util/u_sandbox.c
+++ b/src/xrt/auxiliary/util/u_sandbox.c
@@ -124,6 +124,14 @@ u_sandbox_should_use_ipc(void)
 
 	// Workspace session: app launched by workspace controller with hidden HWND, route to IPC
 	const char *workspace_session = getenv("DISPLAYXR_WORKSPACE_SESSION");
+#ifdef XRT_OS_WINDOWS
+	char workspace_session_buf[16] = {0};
+	if (workspace_session == NULL) {
+		DWORD n = GetEnvironmentVariableA("DISPLAYXR_WORKSPACE_SESSION", workspace_session_buf,
+		                                   sizeof(workspace_session_buf));
+		if (n > 0) workspace_session = workspace_session_buf;
+	}
+#endif
 	if (workspace_session != NULL && strcmp(workspace_session, "1") == 0) {
 		U_LOG_I("DISPLAYXR_WORKSPACE_SESSION=1: forcing IPC mode for workspace controller");
 		return true;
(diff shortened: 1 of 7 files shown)


dfattal (approved):
Tested locally on Windows: 4× gauss splat in shell, 60s bench. PR #201 success criteria met — 2901 fence waits queued across 4 VK clients, stale_views=0, 0 mutex timeouts (legacy path bypassed). Workspace compositor at 60.7 fps p50; demo apps at ~14.6 fps each (separate workspace-throttle issue, not introduced by this PR). Companion demo PR (DisplayXR/displayxr-demo-gaussiansplat#4) also tested and required.

PR #253, opened 2026-05-20
installer: drop system PATH modification; static-link transitive deps

leaiss:
## Summary

The runtime installer no longer modifies system PATH. `DisplayXRClient.dll`'s transitive third-party deps (cjson, SDL2, pthreads) are eliminated at the build layer instead — bundled source for cjson, vcpkg static linkage for pthreads, removed entirely for SDL2 — and the one delay-loaded DLL still co-located in \$INSTDIR (Leia SR Vulkan weaver) is resolved by a DllMain notify hook scoped to delay-loads.

## Why

The old `AddToPath` was needed so Windows could resolve `DisplayXRClient.dll`'s static imports of `pthreadVCE3.dll` / `cjson.dll` / `SDL2.dll`. PATH writes leak across installs — we observed duplicate `C:\Program Files\DisplayXR\Runtime` entries from prior versions on the same machine — and fight with other DisplayXR-adjacent installers. Dropping the PATH write is desirable on its own; eliminating the imports it was masking is the structural fix.

## What changed

- **`vcpkg.json`** — drop `cjson` from deps; drop `gui` from `default-features` (kills SDL2 from the install set; XRT_HAVE_SDL2 auto-detects off).
- **`vcpkg-configuration.json` + `cmake/vcpkg-overlay-triplets/x64-windows.cmake`** (new) — per-port `VCPKG_LIBRARY_LINKAGE` override forces vcpkg's `pthreads` port to build static. `PThreads4WConfig.cmake` auto-detects no `bin/*.dll` and switches the imported target type, so `xrt-pthreads` needs no source change.
- **`src/external/CMakeLists.txt` + `src/xrt/auxiliary/util/u_json.c`** — when the MCP module's `displayxr_mcp_vendored_cjson` static lib is in the build, route `xrt-external-cjson` through it instead of `#include "cjson/cJSON.c"` in `u_json.c`. New `XRT_HAVE_LINKED_CJSON` gate avoids duplicate-symbol collision (`aux_util.lib` vs `displayxr_mcp_vendored_cjson.lib`).
- **`src/xrt/targets/openxr/target_dll_init.c`** (new) — `DllMain` + `__pfnDliNotifyHook2`. Resolves `/DELAYLOAD`'d DLLs from the runtime's own install dir via `LoadLibraryExW + LOAD_WITH_ALTERED_SEARCH_PATH`. Strictly scoped to delay-loads — we're loaded into the host app's process, so `SetDefaultDllDirectories` would be rude.
- **`installer/DisplayXRInstaller.nsi`** — remove `Call AddToPath` in the install section. `un.RemoveFromPath` stays in the uninstaller path so upgrades from older installer versions still clean up stale entries.

## Verification

End-to-end on a Leia SR machine, fresh runtime install with `C:\Program Files\DisplayXR\Runtime` confirmed absent from system PATH:

- `dumpbin /imports DisplayXRClient.dll` — no third-party DLLs in static imports; delay-loaded list correctly shows `SimulatedRealityVulkanBeta.dll` + `vulkan-1.dll`.
- All four cube test apps (D3D11, D3D12, GL, Vulkan) launch and render correctly from a clean install.
- `displayxr-shell.exe` + two cube apps launches, lays out, and chrome renders — paired with a sibling PR in `displayxr-shell-pvt` that mirrors the static-pthreads/cjson change for the shell binary.

## Test plan

- [ ] CI green (build-windows.yml)
- [ ] Smoke: cube_handle_d3d11_win.exe via installed runtime, no PATH workaround
- [ ] Smoke: cube_handle_vk_win.exe via installed runtime (exercises the DllMain delay-load hook for SR weaver)
- [ ] Verify uninstaller still removes legacy PATH entries from installs predating this change

## Release notes / heads-up for downstream consumers

The runtime installer no longer modifies system PATH. Downstream apps that previously relied on `C:\Program Files\DisplayXR\Runtime` being on PATH for any reason need to make their own provisions:

- **OpenXR apps that link the Khronos loader:** bundle `openxr_loader.dll` next to your app's `.exe`. The test apps in this repo (`test_apps/cube_*/CMakeLists.txt`), `displayxr-webxr-bridge`, and the new shell installer all do this via a `POST_BUILD` `copy_if_different` of `OpenXR::openxr_loader`'s `IMPORTED_LOCATION`. Apps that previously found `openxr_loader.dll` in `$INSTDIR` via the PATH entry will need to start shipping their own copy.
- **WebXR consumers:** Chrome's renderer loads `openxr_loader.dll` from the system; this is unchanged. The native WebXR bridge (`displayxr-webxr-bridge.exe`) already co-locates its own loader via the same `POST_BUILD` pattern.
- **Workspace controllers (third-party shells):** statically-imported third-party DLLs need to be self-contained — either statically linked or bundled next to the workspace controller's exe. See the sibling `displayxr-shell-pvt` PR for the canonical pattern (`vcpkg-configuration.json` + per-port static triplet override, plus `install(FILES ... openxr_loader.dll ...)`).

### DIFF
--- a/cmake/vcpkg-overlay-triplets/x64-windows.cmake
+++ b/cmake/vcpkg-overlay-triplets/x64-windows.cmake
@@ -1,0 +1,17 @@
+set(VCPKG_TARGET_ARCHITECTURE x64)
+set(VCPKG_CRT_LINKAGE dynamic)
+set(VCPKG_LIBRARY_LINKAGE dynamic)
+
+# Force pthreads to build as a static library so DisplayXRClient.dll has no
+# runtime dependency on pthreadVCE3.dll. The PThreads4WConfig.cmake provided
+# by the port auto-detects the absence of bin/*.dll and creates an UNKNOWN
+# IMPORTED CMake target — same target name (PThreads4W::PThreads4W_CXXEXC),
+# so xrt-pthreads requires no source changes.
+#
+# Static pthreads is the only piece statically linked from vcpkg; everything
+# else (Vulkan loader, glslang, eigen) keeps its default linkage. This lets
+# the installer drop its system-PATH modification (issue: PATH was previously
+# required to resolve pthreadVCE3.dll alongside DisplayXRClient.dll).
+if(PORT STREQUAL "pthreads")
+    set(VCPKG_LIBRARY_LINKAGE static)
+endif()
--- a/installer/DisplayXRInstaller.nsi
+++ b/installer/DisplayXRInstaller.nsi
@@ -662,16 +662,26 @@ Section "DisplayXR Runtime" SecRuntime
 	; SetRegView; explicit no-op kept for clarity).
 	WriteRegStr HKLM "Software\Khronos\OpenXR\1" "ActiveRuntime" "$INSTDIR\DisplayXR_win64.json"
 
-	; PATH is intentionally not modified. DisplayXRClient.dll has no static
-	; DLL imports that live in $INSTDIR — cjson is compiled in from bundled
-	; source, SDL2 is disabled in the production build, and pthreads is
-	; statically linked via the vcpkg overlay triplet at
-	; cmake/vcpkg-overlay-triplets/x64-windows.cmake. The only DLL co-located
-	; in $INSTDIR (SimulatedRealityVulkanBeta.dll) is /DELAYLOAD'd and only
-	; reached by the Vulkan compositor path, which is loaded via LoadLibrary
-	; from the service exe's own directory (already on the default search path).
-	; The uninstaller still runs un.RemoveFromPath to clean up entries written
-	; by older installer versions.
+	; PATH is intentionally not modified. DisplayXRClient.dll's third-party
+	; transitive deps that previously *required* $INSTDIR on PATH are gone:
+	; cjson is compiled in from bundled source, SDL2 is disabled in the
+	; production build, and pthreads is statically linked via the vcpkg
+	; overlay triplet at cmake/vcpkg-overlay-triplets/x64-windows.cmake.
+	;
+	; $INSTDIR still receives other files via the wildcard at the File line
+	; above (openxr_loader.dll, the five static-imported SimulatedReality*.dll
+	; shipped by drv_leia's import-lib link, etc.), but none of those need
+	; PATH to resolve:
+	;  - SimulatedReality{Core,Displays,FaceTrackers,DirectX,OpenGL}.dll are
+	;    also installed by the Leia SR Platform (a hard prereq of drv_leia),
+	;    which adds its own dir to system PATH independently — static imports
+	;    resolve through that.
+	;  - SimulatedRealityVulkanBeta.dll is /DELAYLOAD'd and resolved by
+	;    DisplayXRClient.dll's __pfnDliNotifyHook2 (target_dll_init.c), which
+	;    LoadLibraryEx's it from this $INSTDIR at first use.
+	;
+	; The uninstaller still runs un.RemoveFromPath to clean up entries
+	; written by older installer versions.
 
 	; Enable D3D11 native compositor by default
 	; This bypasses Vulkan and avoids D3D11<->Vulkan interop issues on Intel GPUs
--- a/installer/DisplayXRInstaller.nsi
+++ b/installer/DisplayXRInstaller.nsi
@@ -662,11 +662,16 @@ Section "DisplayXR Runtime" SecRuntime
 	; SetRegView; explicit no-op kept for clarity).
 	WriteRegStr HKLM "Software\Khronos\OpenXR\1" "ActiveRuntime" "$INSTDIR\DisplayXR_win64.json"
 
-	; Add install directory to system PATH
-	; This is needed so OpenXR apps can find DisplayXRClient.dll's dependencies
-	; (vulkan-1.dll, SDL2.dll, etc.) when loading the runtime
-	Push $INSTDIR
-	Call AddToPath
+	; PATH is intentionally not modified. DisplayXRClient.dll has no static
+	; DLL imports that live in $INSTDIR — cjson is compiled in from bundled
+	; source, SDL2 is disabled in the production build, and pthreads is
+	; statically linked via the vcpkg overlay triplet at
+	; cmake/vcpkg-overlay-triplets/x64-windows.cmake. The only DLL co-located
+	; in $INSTDIR (SimulatedRealityVulkanBeta.dll) is /DELAYLOAD'd and only
+	; reached by the Vulkan compositor path, which is loaded via LoadLibrary
+	; from the service exe's own directory (already on the default search path).
+	; The uninstaller still runs un.RemoveFromPath to clean up entries written
+	; by older installer versions.
 
 	; Enable D3D11 native compositor by default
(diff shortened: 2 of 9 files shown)


dfattal:
Flagging a potential gap before this lands — the PR description's verification claim doesn't match the runtime as currently linked. Worth re-checking on a clean box.

## Static `SimulatedReality*.dll` imports aren't covered by the delay-load hook

`dumpbin /imports _package\bin\DisplayXRClient.dll` (pre-PR build) shows these as **static (non-delay) imports** that live in `\$INSTDIR`:

```
SimulatedRealityCore.dll
SimulatedRealityDisplays.dll
SimulatedRealityFaceTrackers.dll
SimulatedRealityDirectX.dll       (D3D11 + D3D12 weaver paths)
SimulatedRealityOpenGL.dll        (GL weaver path)
```

These come from `drv_leia` linking the corresponding import libs in `src/xrt/drivers/CMakeLists.txt:148-214`. The PR's `__pfnDliNotifyHook2` **only fires for `/DELAYLOAD`'d DLLs**, and `src/xrt/targets/openxr/CMakeLists.txt:26-29` only delay-loads `vulkan-1.dll` + `SimulatedRealityVulkanBeta.dll`. So the new hook can't help these five.

The Khronos loader `LoadLibrary`s `DisplayXRClient.dll` via the absolute path in `DisplayXR_win64.json`. At that moment Windows resolves DisplayXRClient's static imports through the **host app's** search order: AppDir → System32 → CurrentDir → PATH. With PATH removed and these DLLs only in `\$INSTDIR`, **DisplayXRClient.dll fails to load** for any OpenXR app whose AppDir doesn't already contain copies of the five SR DLLs. None of our cube test apps ship them.

## Why this likely didn't surface in verification

The auto-generated dev run scripts still prepend `_package\bin` to PATH (`scripts/build_windows.bat:305`, and the resulting `_package/run_cube_handle_vk_win.bat` etc.). Local smoke tests via those scripts will keep working even when the installed runtime is broken, because dev PATH masks the missing static imports. To repro the production failure, run from a vanilla `cmd.exe`:

```cmd
set PATH=C:\Windows\System32;C:\Windows
\"C:\path\to\installed\app\openxr_app.exe\"
```

…or just `dumpbin /imports` the PR-built DLL and confirm the static SR imports are gone.

## Suggested fixes

Either:
- Add the five SR DLLs to `/DELAYLOAD` so the new hook resolves them from `\$INSTDIR` (cheapest), or
- Switch `drv_leia` to runtime `LoadLibrary` + `GetProcAddress` for SR entry points instead of import-lib linkage.

## Minor stuff while you're in there

- Stale comment in `test_apps/cube_handle_vk_win/CMakeLists.txt:131-134` (and `cube_hosted_legacy_vk_win/CMakeLists.txt:124`) still references the PATH mechanism.
- Installer comment at `installer/DisplayXRInstaller.nsi:666-672` says *\"the only DLL co-located in \$INSTDIR is SimulatedRealityVulkanBeta.dll\"* — \$INSTDIR actually receives all five SR core DLLs + `openxr_loader.dll` + `pthreadVC3.dll` etc. via the wildcard at line 652.
- Third-party / WebXR-launched apps that used to find `\$INSTDIR\openxr_loader.dll` via PATH will need to start shipping their own loader. Worth a line in the release notes.

Net: pthreads/cjson/SDL2 cleanup is solid, direction is right — just need to either delay-load or runtime-load the SR core libs too before the PATH write can come out.

dfattal:
Correction — verified on a Leia SR dev machine that the SR Platform installer puts `C:\Program Files\LeiaSR\Platform\bin` on the **machine-level system PATH**, and that directory contains all five `SimulatedReality{Core,Displays,FaceTrackers,DirectX,OpenGL}.dll`. Since SR Platform is a hard prereq of `drv_leia` (it's import-lib linked, runtime can't load without it), those static imports will resolve through SR Platform's own PATH entry — independent of whether `$INSTDIR` is on PATH.

So my flagged concern is a non-issue in practice. The new DllMain hook still does the right thing for the one DLL that *isn't* shipped by SR Platform (`SimulatedRealityVulkanBeta.dll` — the runtime ships its own fallback copy, see `src/xrt/drivers/CMakeLists.txt:170-175`).

Sorry for the noise. The minor items (stale comments, third-party loader-bundling note) still stand if you want to clean them up.

leaiss:
Thanks for the thorough review — and for the self-correction on the static-import resolution. Agreed the concern is moot in practice today (SR Platform's PATH entry resolves the five static SR imports independently of `$INSTDIR`), so the PR as-is doesn't regress that path.

Following up on the suggestion in your first comment though — adding the five SR DLLs to `/DELAYLOAD` — I want to record a reservation here even though we don't *need* to act on it in this PR.

## Reservation: hardcoded vendor DLL names contradict the vendor-agnostic design intent

`CLAUDE.md` and the architecture docs frame DisplayXR as vendor-agnostic — *"any 3D display vendor can integrate their driver. Leia SR SDK is the first vendor integration."* The current Windows build is de-facto Leia-coupled because `drv_leia` import-lib-links five `SimulatedReality*.dll`s directly into `DisplayXRClient.dll`. Adding those same five names to `/DELAYLOAD` would propagate Leia-specific identifiers from the driver layer (where they belong) up into the runtime target's link line (where they don't). It hardens a vendor-coupled shape into the *runtime* target, which is the wrong layer to know about Leia.

What the `/DELAYLOAD` suggestion would actually buy us today, and what it wouldn't:

- ✅ Theoretical seam: if someone later bundles SR DLLs in `$INSTDIR`, the existing `__pfnDliNotifyHook2` would auto-pick them up.
- ❌ Today: the hook tries `$INSTDIR\SimulatedRealityCore.dll`, returns NULL (file not present), and the helper falls back to default search → still resolves via SR Platform's PATH. So no observable behavior change.
- ❌ Vendor isolation: doesn't actually achieve it. The runtime DLL still has "SimulatedReality*" hardcoded in the linker arguments. A different display vendor's runtime build would either need to delete these `/DELAYLOAD` entries or have the build system invent its own list.

The principled fix for vendor isolation is the **vendor-plug-in restructure** — `drv_leia` becomes a separately-built module that DisplayXR `LoadLibrary`s at runtime via the display-processor vtable contract in `docs/guides/vendor-integration.md`. Then `DisplayXRClient.dll` has *zero* mentions of any vendor's DLL names, static or delay-loaded, and the four display-processor variants (D3D11, D3D12, GL, Vulkan, Metal) become discoverable plug-ins. That's the layer this work belongs at.

So: declining the `/DELAYLOAD` suggestion for this PR. Filing the vendor-plug-in restructure as the proper follow-up rather than chasing it tactically here.

## Minor items I'll roll into this PR

Picking these up since I'm in the area:

- Stale `_package\bin` PATH comments in `test_apps/cube_handle_vk_win/CMakeLists.txt:131-134` and `cube_hosted_legacy_vk_win/CMakeLists.txt:124`.
- Outdated `$INSTDIR` co-location comment in `installer/DisplayXRInstaller.nsi:666-672` (the wildcard at line 652 sweeps in more than just `SimulatedRealityVulkanBeta.dll`).
- Release-note bullet for third-party / WebXR apps that previously found `openxr_loader.dll` via the runtime's PATH entry — they need to bundle their own loader (matches the pattern test apps + the shell installer already follow).

dfattal (approved):
good will merge this then rebase the leia / sim display plugin refactor on top

Issue #311, opened 2026-05-24
displayxr-demo-gaussiansplat: ship a macOS .pkg


dfattal:
The Gaussian-splat demo at [DisplayXR/displayxr-demo-gaussiansplat](https://github.com/DisplayXR/displayxr-demo-gaussiansplat) ships Windows-only today (`DisplayXRGaussianSplatSetup-1.3.2.exe`). The macOS source (`macos/main.mm`, `macos/CMakeLists.txt`, `scripts/build_macos.sh`) already produces a working flat CLI binary, but there is no packaging.

## What this issue covers

Add a macOS `.pkg` shipping path to `displayxr-demo-gaussiansplat`:

- `installer/macos/build_installer.sh` + `create_app_bundle.sh` + `Distribution.xml` + `resources/` (LICENSE, welcome.html)
- `.github/workflows/build-macos.yml` — tag-triggered (`v*`) macOS CI that produces and attaches `DisplayXRGaussianSplat-<version>.pkg` to the GitHub Release via `softprops/action-gh-release@v2`
- Extend `scripts/build_macos.sh` with an `--installer` flag (mirroring the runtime's pattern)

Tag the first multi-platform release as `v1.4.0`.

## Why

The meta-installer bundle ([displayxr-installer](https://github.com/DisplayXR/displayxr-installer), #284) bundles 5 components on Windows but only the runtime on macOS — components without a macOS `.pkg` are silently skipped at bundle-build time. Adding a Gauss `.pkg` makes it the first demo to join the macOS bundle and exercises the bundle's graceful-degradation path in the "add a new component" direction.

## Patterns to reuse

- `installer/macos/build_installer.sh` (pkgbuild + productbuild + ad-hoc codesign rule from PR #279)
- `installer/macos/create_app_bundle.sh` (.app bundle assembly with launcher + bundled dylibs)
- `.github/workflows/build-macos.yml` BuildInstaller job
- The .pkg expects the DisplayXR runtime .pkg to be installed first (mirrors how Windows orders runtime → demo); it bundles `openxr_loader.1.dylib` + `libvulkan.1.dylib` + `libMoltenVK.dylib` inside the .app but defers to `/etc/xdg/openxr/1/active_runtime.json` for runtime discovery.

## Follow-up here

[...]

Issue #284, opened 2026-05-22
User meta-installer: one-download full-stack DisplayXR for end users (new displayxr-installer repo)


dfattal:
Today an end user who wants "DisplayXR to just work" has to download + run 4-5 separate installers (runtime, Shell, Leia plug-in, MCP Tools, plus per-demo). That's appropriate for the current developer-facing audience but won't scale to end-user distribution. This issue tracks a single-download meta-installer.

> **Companion issue**: see #283 for the **developer-facing orchestrator script**. Both share `versions.json` semantics — the schema is defined once and consumed in both places.

## Architectural shape

Both Windows and macOS have native primitives for chained installs:

- **Windows**: NSIS supports `ExecWait '$INSTDIR\nested.exe /S'` for chaining component installers. The bundle is either an offline `.exe` that drops + runs each component, or a bootstrapper that downloads them from GitHub Releases at install time. NSIS runs the bundle with `RequestExecutionLevel admin`, so children inherit elevation and don't re-prompt UAC.
- **macOS**: `productbuild --distribution Distribution.xml` is *literally designed* for this — list each component `.pkg` as a `<pkg-ref>` and the installer GUI shows checkboxes for optional components. We already use this pattern in `installer/macos/Distribution.xml` for the runtime + cube test app; extending it to include Shell + plug-ins is the same shape. Embedded child pkgs are read internally by `Installer.app` and don't quarantine separately.

## Where it lives — **new `DisplayXR/displayxr-installer` public repo**

NOT in `displayxr-runtime/releases`. Three reasons:

1. **Branding clarity.** A developer integrating DisplayXR into their app wants `DisplayXRClient.dll` — they should not get a 200 MB bundle of Shell + demos when they `gh release download`. "Runtime release" and "DisplayXR release" need to be different artifacts on different release pages.
2. **Independent cadence.** The bundle ships when the *whole stack* is ready (launch-driven). The runtime ships when there's a runtime fix worth tagging (engineering-driven). Coupling them forces one team's cadence onto the other.
3. **Compat matrix lives somewhere honest.** "runtime v1.4.0 pairs with Shell v0.9.1 pairs with Leia plug-in v1.2.0" gets pinned and tested as a unit. That belongs in a dedicated repo with `versions.json` + a small CI matrix, not nested inside the runtime's release notes.

## Repo layout

```
DisplayXR/displayxr-installer                   ← new public repo
[...]

Issue #283, opened 2026-05-22
Dev orchestrator: one-command full-stack DisplayXR install for contributors


dfattal:
Today a contributor who wants a working DisplayXR dev box has to manually orchestrate ~5 separate installers (runtime, Shell, Leia plug-in, MCP Tools, optionally each demo) — clone, build/download, install, repeat. None of the architectural pieces block a single-command setup: the runtime + Shell + plug-ins already coordinate via the registry/manifest discovery contract we just shipped on macOS, and every artifact except the Shell source is public.

This issue tracks a top-level orchestrator that compresses the setup into one command.

> **Companion issue**: see #284 (user meta-installer) for the end-user side. Both share `versions.json` as the source of truth for the cross-repo compat matrix; this dev orchestrator and the user bundle should agree on that schema from day 1.

## Scope

Two layers, gated by a flag:

### Binary-bundle path (default, ~5 min, no SDK needed beyond runtime's own)

A new `scripts/setup-displayxr.sh` (and `scripts\setup-displayxr.bat`) that:

1. Reads a new top-level `versions.json` pinning the cross-repo compat matrix:
   ```json
   {
     "runtime": "v1.4.0",
     "shell": "v0.9.0",
     "leia": "v1.2.0",
     "mcp": "v0.3.0"
   }
   ```
2. `gh release download` each component's installer per platform.
3. Silent-installs each (`installer /S` on Windows; `sudo installer -pkg ... -target /` on macOS).
4. Default install set: runtime + Shell + Leia plug-in + MCP Tools.
5. Optional flags:
   - `--with demos` — also clones each `displayxr-demo-*` repo into a sibling `demos/<name>/` and builds it.
   - `--without shell,leia,mcp` — opt out of components for minimal installs.

Repos consumed (all public; Shell-source is private but its **public installer release** at `displayxr-shell-releases` is what the orchestrator pulls):

| Component | Repo it reads from | Auth |
|---|---|---|
| Runtime | `DisplayXR/displayxr-runtime` (this repo) | none |
| Sim-display plug-in | (bundled in runtime installer) | none |
| DisplayXR Shell | `DisplayXR/displayxr-shell-releases` (public installer mirror) | none |
| Leia SR plug-in | `DisplayXR/displayxr-leia` (post-`feat/extract-leia-plugin`) | none |
| MCP Tools | `DisplayXR/displayxr-mcp` | none |
| Demos | `DisplayXR/displayxr-demo-*` | none |

[...]

PR #339, opened 2026-05-26
Orchestrator --with-demos installs demos; --with-demo-sources clones (#297)

### LINKED ISSUE #297: Orchestrator --with-demos should also install demos (not just clone source)
## Today

`scripts/setup-displayxr.{sh,bat} --with-demos` only **clones** each `displayxr-demo-*` repo into `./demos/<name>/`. It does **not** install. That was intentional in #283 because demo builds were API-specific and pre-#284 most demos didn't have a one-line installer to run.

## What changed

Demos with `.exe` / `.pkg` releases now exist — `displayxr-demo-gaussiansplat` v1.3.1 ships `DisplayXRGaussianSplatSetup-1.3.1.exe` today, and the meta-installer scope from #284 will pin one or more demos into the bundle. For those demos, the orchestrator could trivially `gh release download` + `start /wait "" $exe /S` exactly like it does for runtime / shell / leia / mcp — that table of `(repo, exe-glob, marker-key)` is already the right shape.

## Why this matters now

On a fresh dev box doing `setup-displayxr.bat --with-demos`:

- Today: clones to `./demos/displayxr-demo-gaussiansplat/`, but the shell launcher shows no Gaussian Splat tile and the demo isn't usable. The contributor has to find the `.exe` on the demo's release page and double-click it separately — not a one-command setup.
- With this change: every pinned demo with a release asset is installed silently in the same pass as the runtime/shell/leia/mcp components. The shell launcher's manifest scan picks them up on next launch.

## Proposed shape

1. **`versions.json`**: already has `demos: { gaussiansplat: "v1.3.1" }`. Reuse as-is.
2. **`scripts/lib/components.sh`**: add `COMPONENT_REPO_demo_<name>` / `COMPONENT_EXE_demo_<name>` entries for each demo with a release. Asset glob is the per-demo `DisplayXR<Demo>Setup-*.exe`.
3. **`scripts/setup-displayxr.bat` + `.sh`**: when `--with-demos` is set (or a new dedicated `--install-demos` flag), iterate `versions.json`'s `demos` map and call the existing `install_component` subroutine for each. Same elevation / asset-attached / marker-check semantics — zero new code path.
4. **Backwards compat**: keep the existing `gh repo clone` behavior either as a separate `--with-demo-sources` flag or as a sibling of `--install-demos`. Contributors building from source still want the clone.

## Coordination with #284

This is **probably the wrong place for it long-term** — once #284's `displayxr-installer` meta-installer ships, end users will get the demos via the bundle, not via the dev orchestrator. The orchestrator is dev-facing.

But the orchestrator is *also* a dev's only one-command path until #284 lands, and #284 doesn't have an ETA. Two ways to think about it:

- **(A) Land in the orchestrator now, retire when #284 ships.** Cheap (~30 line diff per OS). Bridges the gap; one place for a contributor to discover what "full stack" means today. When #284 ships, deprecate the orchestrator-side demos in favor of pointing at the bundle.
- **(B) Wait for #284 and don't bloat the orchestrator.** Keeps the dev script narrowly scoped at runtime/shell/leia/mcp. Contributors continue to install demos manually until the bundle ships.

I'd weakly recommend **(A)** if #284 is more than ~1 month out; **(B)** if it's imminent.

## Why this is filed against the runtime repo

The orchestrator lives in `displayxr-runtime/scripts/`. Filing here keeps the implementation discussion next to the code. Linked to #284 since the user-facing bundle work supersedes this in the long run.

## Surfaced by

Tail-of-the-week e2e test on DisplayXR/displayxr-runtime#291. After `setup-displayxr.bat --with mcp` the shell launcher had no Gauss tile until I separately ran the demo installer. See DisplayXR/displayxr-demo-gaussiansplat#10 for the related demo-side uninstaller fragility found in the same trace.

Linked: #284

dfattal:
Closes #297.

## What

Splits the dev orchestrator's demo handling into two intent-revealing flags:

| Flag | Behavior |
|---|---|
| `--with-demos` | **Installs** each demo's prebuilt release asset — `gh release download` + silent install + marker check, the *same* `install_component` path as runtime/shell/leia/mcp. |
| `--with-demo-sources` | **Clones** each `displayxr-demo-*` repo into `./demos/<name>/` (the former `--with-demos` behavior), for contributors who build from source. |

Both combine freely; `--dry-run` works with both.

## Why (the resilience argument)

The original clone-only choice (#283) was to avoid importing each demo's **build** environment (Vulkan SDK, graphics toolchains) onto the contributor's box. But clone-only doesn't make a demo *usable* — it leaves the dev to build it, which is exactly where that fragility bites; after `--with-demos` the shell launcher showed no demo tile (the bug #297 was filed for).

Installing a **prebuilt** release asset sidesteps the build entirely — the binary was already compiled by the demo's own CI — so `--with-demos` is *more* resilient to dev-environment drift, not less. Source-only demos (no release installer) stay under `--with-demo-sources`.

## How

- `components.sh`: new `COMPONENT_PIN_KEY_<name>` override (the demo's tag lives at `demos.gaussiansplat`, not a top-level key) + a `DEMO_COMPONENTS` list. `install_component` resolves the pin key, defaulting to the component name.
- `setup-displayxr.sh` / `.bat`: `--with-demos` now loops `DEMO_COMPONENTS` through `install_component` (after the core components, so runtime is present for demos that require it); `--with-demo-sources` carries the old clone loop. Demos with no asset for the current OS warn-and-skip (no hard failure).
- `full-stack-install.md`: documents the split.

#311 already registered `gauss_demo` in the component table; this PR wires it into the install path. Bridges the gap until #284's user-facing meta-installer ships.

## Test

- `bash -n` clean on `.sh` + `components.sh`; `--help` renders the new flags.
- Pin-key resolution verified: `runtime` → `runtime`, `gauss_demo` → `demos.gaussiansplat` → `v1.4.1`.
- `.bat` mirrors the same structure (exact-match arg parse; `--with-demos` doesn't shadow `--with-demo-sources`).

### DIFF
--- a/docs/getting-started/full-stack-install.md
+++ b/docs/getting-started/full-stack-install.md
@@ -42,13 +42,15 @@ OpenXR runtime via `HKLM\Software\Khronos\OpenXR\1\ActiveRuntime`.
 |---|---|
 | _(none)_ | macOS: runtime only. Windows: runtime + Shell + Leia plug-in. |
 | `--with mcp` | Also install DisplayXR MCP Tools (now dual-platform: `DisplayXRMCPSetup-*.exe` on Windows, `DisplayXRMCP-*.pkg` on macOS). |
-| `--with-demos` | Clone each `DisplayXR/displayxr-demo-*` repo into `./demos/<name>/`. Does not build them. |
+| `--with-demos` | Also **install** each demo's prebuilt release asset (same download → silent-install → marker-check path as the runtime — no source build, so a contributor's graphics toolchain / Vulkan SDK is irrelevant). A demo with no release asset for this OS warn-and-skips. |
+| `--with-demo-sources` | **Clone** each `DisplayXR/displayxr-demo-*` repo into `./demos/<name>/` for building from source. Does not build them — see each demo's README. Independent of `--with-demos`. |
 | `--dry-run` | Print every download / install command without running it. |
 | `--uninstall` | macOS: chain `/Library/Application Support/DisplayXR/uninstall.sh`. Windows: silent-uninstall every `Publisher=DisplayXR` component. |
 | `-h`, `--help` | Show usage. |
 
-`--with` and `--with-demos` combine freely. `--dry-run` works with all
-other flags.
+`--with`, `--with-demos`, and `--with-demo-sources` combine freely (pass
+both demo flags to install the released demos *and* clone their sources).
+`--dry-run` works with all other flags.
 
 ## Prerequisites
 
@@ -90,9 +92,9 @@ Top-level pin file. Bump a value here to roll the dev box forward:
   "shell":       "v1.2.5",
   "leia_plugin": "v1.0.3",
   "mcp_tools":   "v0.3.2",
-  "demos": {
-    "gaussiansplat": "v1.4.1"
-  }
+  "gauss_demo":  "v1.4.1",
+  "modelviewer_demo": "v0.7.0",
+  "mediaplayer_demo": "v1.0.0"
 }
 ```
 
--- a/scripts/lib/components.sh
+++ b/scripts/lib/components.sh
@@ -14,9 +14,19 @@
 #                                             successful install on macOS. Empty = skip check.
 #   COMPONENT_INSTALL_MARKER_WINDOWS_<name>   Registry key whose existence proves a
 #                                             successful install on Windows. Empty = skip check.
+#   COMPONENT_PIN_KEY_<name>                  versions.json key holding this component's tag,
+#                                             when it differs from <name> (e.g. nested under
+#                                             "demos"). Empty = the pin key equals <name>.
 #
 # A blank platform glob means "warn-and-skip on this platform" — used today
 # for shell/leia/mcp on macOS.
+#
+# Demos are *installed from their prebuilt release asset*, exactly like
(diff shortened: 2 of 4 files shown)


dfattal:
**Validated with a real macOS install** (`./scripts/setup-displayxr.sh --with-demos`):

- Runtime `.pkg` installed → `installer: The upgrade was successful` → ✅ `runtime installed`.
- `--with-demos` resolved `gauss_demo` via the new `demos.gaussiansplat` pin key → `v1.4.1`, downloaded `DisplayXRGaussianSplat-1.4.1.pkg` → ✅ `gauss_demo installed`.
- Smoke verification green; all three markers confirmed present afterward: runtime sim-display manifest, `/etc/xdg/openxr/1/active_runtime.json`, and `/Applications/Gaussian Splat Viewer.app`.

The run also surfaced a latent robustness bug (pre-existing in `install_component`, now exercised by the demo path): a transient `gh` broken-pipe on the asset-count query left `asset_count` empty and `[ "$asset_count" -eq 0 ]` aborted with `integer expected`. It happened to fall through and install anyway, but that was accidental. Fixed in `8cc17e93d` — the integer test is now guarded by a numeric match, so a network blip falls through to the download (the real gate) cleanly instead of printing a scary error.

PR #340, opened 2026-05-27
docs: refresh emulator setup doc post-merge (broker + namespace fix landed)

leaiss:
## Summary

PR #331 (the emulator setup doc) merged before PRs #332 (runtime broker) and #333 (plug-in namespace fix) — so as soon as those two landed, the doc became out of date. This PR refreshes it against the actual current main.

Caught by running a fresh-checkout dry-run of the merged docs against the merged code.

## What changes

1. **\"What works\" table** — broker discovery, plug-in dlopen with the namespace preload fix, plug-in probe + negotiate, and \`xrCreateInstance\` now all work end-to-end on the emulator. Previously this doc framed the broker as the wall.

2. **\"What doesn't work\" table** — the actual wall is now \`xrCreateVulkanDeviceKHR\` returning \`VK_ERROR_EXTENSION_NOT_PRESENT\` (swiftshader missing extensions the runtime requests). Lume Pad's hardware Vulkan should clear that.

3. **Build procedure (step 2) was wrong in 3 ways** for a fresh checkout:
   - **Order**: plug-in first then runtime — but the plug-in CMake consumes \`Eigen3_DIR\` from the runtime's build/intermediates/ tree, which only exists after the runtime build runs once. Reversed: runtime first, then plug-in.
   - **Bundle**: showed basic \`scripts/build-android.sh\` + single \`cp\` of the plug-in .so. Produces an APK with 2 .so files instead of 16; runtime fails at first \`xrCreateInstance\` with \`dlopen libleiaSDK-faceTrackingInApp.so not found\`. Switch to \`scripts/build-android.sh install-runtime-jnilibs\` (PR plug-in #8 — pending merge) which bundles all 16 .so files in one command.
   - **Re-build step**: the third runtime build (\`--rerun-tasks\` so gradle picks up the freshly-installed jniLibs/<ABI>/) was implicit. Now explicit.

4. **Expected-log section** updated to show the full success path through broker → preload → plug-in negotiate → \`xrCreateInstance\` → \`xrCreateVulkanInstanceKHR\` → \`xrCreateVulkanDeviceKHR vk_result=-7\`.

## Sanity check for users

Added to the doc: \`unzip -l <runtime.apk> | grep '\.so\$'\` should list 16 entries after a correct build. If it shows only 1 or 2, the \`install-runtime-jnilibs\` step didn't run or the runtime rebuild didn't pick them up.

## Dependency

The new build procedure relies on the plug-in repo's \`scripts/build-android.sh install-runtime-jnilibs\` target — currently in [leia-plugin#8](https://github.com/DisplayXR/displayxr-leia-plugin/pull/8), pending merge. Once that lands the doc commands work as written.

### DIFF
--- a/docs/getting-started/android-emulator-setup.md
+++ b/docs/getting-started/android-emulator-setup.md
@@ -1,10 +1,12 @@
 # Android emulator pre-bring-up
 
 How far the Android emulator gets you on the way to a real Lume Pad
-bring-up. Short version: **build + install + most of the OpenXR loader
-discovery surface — but xrCreateInstance fails at the broker step.**
-Useful for catching half the bring-up bugs early; not a full
-replacement for hardware.
+bring-up. Short version: **build + install + the full OpenXR
+loader / broker / plug-in handshake all work — and
+`xrCreateInstance` succeeds.** The wall is at the runtime's own
+Vulkan-device creation, where the emulator's software-Vulkan is
+missing extensions the runtime asks for. Useful for catching every
+build / loader / linker bug before hardware.
 
 ## What works on the emulator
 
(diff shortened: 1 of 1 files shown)


dfattal:
Stale — predates #498. Emulator/bring-up docs should be refreshed against the current Android state. Closing.

PR #347, opened 2026-05-27
android: smoke-test sentinel + local runner script

leaiss:
## Summary

Stacked on PR #343 (preload fix). Adds two things to make the runtime+plug-in regression that #343 fixed catchable going forward:

1. **`ANDROID_POC_SENTINEL`** log line in `cube_handle_vk_android` immediately after `xrCreateInstance -> XR_SUCCESS`. Distinctive string a single grep can match.
2. **`scripts/android-smoketest.sh`** — single-command local validation: boots an emulator if no device attached, builds the leia-plugin + drops 16 transitive `.so` into the runtime APK's `jniLibs/`, builds runtime + test APKs, installs both, launches the test app, polls logcat for the sentinel with a 60s timeout.

Tested locally end-to-end on an Android-36 x86_64 emulator: passes in ~6s.

## Why local instead of CI

I originally tried adding a CI emulator gate to `build-android.yml`, but cross-org CNSDK fetch from `LeiaInc/leiainc.github.io` is wedged:

- Workflow `GITHUB_TOKEN` is App-scoped to `DisplayXR` → 404 against LeiaInc repos via either REST or git+token.
- Anonymous git clone is rejected — github.com requires auth, and the runner's credential helper turns the auth prompt into a hard fail (`GIT_TERMINAL_PROMPT=0`).
- LFS raw URLs 404.

Three commit attempts (sparse LFS checkout, token clone, GIT_ASKPASS=/bin/true) all failed at different layers. None is a 5-min fix — they all need either a PAT secret with cross-org read or vendoring CNSDK into a DisplayXR repo. Both are doable later; neither is needed to ship regression-catching now.

## Usage

```bash
# From the runtime repo root:
scripts/android-smoketest.sh

# Common env overrides:
PLUGIN_DIR=/c/displayxr-leia-plugin \
CNSDK_ROOT=./cnsdk \
SKIP_EMULATOR=1 \
scripts/android-smoketest.sh
```

Exits 0 on pass, 1 on hard fail (FATAL / XR_ERROR), 2 on timeout, 3 on missing prereq.

## Stacking

Based on PR #343. Rebase onto main after #343 merges.

### DIFF
--- a/scripts/android-smoketest.sh
+++ b/scripts/android-smoketest.sh
@@ -142,6 +142,20 @@ build_plugin_and_install_jnilibs() {
     echo "[smoketest] building plug-in + installing transitive jniLibs"
     (
         cd "$PLUGIN_DIR"
+        # Honor PLUGIN_BRANCH (documented in the header): check it out before
+        # building so the smoke test builds a known plug-in revision. Requires a
+        # clean plug-in working tree; abort loudly rather than build a stale tip.
+        if [[ -n "${PLUGIN_BRANCH:-}" ]]; then
+            if [[ -n "$(git status --porcelain 2>/dev/null)" ]]; then
+                echo "ERROR: PLUGIN_BRANCH set but $PLUGIN_DIR has uncommitted changes" >&2
+                exit 3
+            fi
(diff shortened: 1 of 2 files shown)


dfattal:

PR #350, opened 2026-05-27
android(test app): declare CAMERA + WAKE_LOCK permissions for Lume Pad

leaiss:
## Summary

Pre-Lume-Pad audit of the test app's `AndroidManifest.xml`. Adds two missing permissions that would otherwise silently degrade the day-1 experience, plus a Step 0 in `android-bringup-checklist.md` documenting how to grant them.

## What this fixes (silent failures avoided)

| Missing | Day-1 symptom | Fix |
|---|---|---|
| `android.permission.CAMERA` | CNSDK's `libleiaSDK-faceTrackingInApp.so` opens the front camera at `leia_core_enable_face_tracking`. Without CAMERA: silent open() failure, no face reported, `xrLocateViews` always returns default centered eye position. **No crash, no log error** — just no head tracking. Hours of "why isn't head tracking working" debugging. | Declared in manifest + `pm grant` documented |
| `uses-feature camera.front` | Play Store filters the app off no-camera devices, AND Android may not surface camera in runtime dialog | Declared `required="false"` so emulator (no camera) still installs |
| `android.permission.WAKE_LOCK` | Long XR sessions don't generate input → screen dims/sleeps → session pauses | Declared (install-time grant) |

## Notes on CAMERA runtime grant

CAMERA is "dangerous" on Android 6+; declaring it in the manifest isn't enough — the user must grant at runtime. The test app is a pure `NativeActivity` with no Java/Kotlin wrapper to surface the system dialog. Two workarounds documented in the bring-up checklist:

1. **Pre-grant via adb** (works on emulator + developer-mode device):
   ```bash
   adb shell pm grant com.displayxr.cube_handle_vk_android android.permission.CAMERA
   ```
2. **Wait for CNSDK's own dialog** at first `leia_core_enable_face_tracking` — depends on CNSDK behavior on Lume Pad, untested pre-hardware.

Production-quality permission handling (wrap `NativeActivity` in a Kotlin `Activity` that calls `requestPermissions()` at `onCreate`) deferred — that's a separate ~50 LOC change with its own review concerns.

## Verification

Built + installed on the Android-36 emulator:

```
$ adb shell dumpsys package com.displayxr.cube_handle_vk_android | grep -E 'CAMERA|WAKE_LOCK|INTERNET'
      android.permission.INTERNET
      android.permission.CAMERA
      android.permission.WAKE_LOCK
      android.permission.INTERNET: granted=true
      android.permission.WAKE_LOCK: granted=true
[...]

PR #359, opened 2026-05-28
android(test app): MainActivity wrapper requests CAMERA at onCreate

leaiss:
## Summary

Pick-up of the deferred follow-up from #350. Adds a thin Kotlin `MainActivity` subclass of `NativeActivity` whose only job is to call `requestPermissions(CAMERA)` from `onCreate`. Stacked on top of #350 (the manifest declarations the wrapper depends on).

## Why this matters

CNSDK's face tracker (`libleiaSDK-faceTrackingInApp.so`, loaded into the app's process by the DisplayXR runtime broker) opens the front camera during `xrCreateSession`. On Android 6+ a manifest `<uses-permission>` declaration is **not sufficient** for "dangerous" perms — the user must explicitly grant at runtime. Without the grant: camera `open()` silently fails, CNSDK never reports a face, all `xrLocateViews` calls return the default centered eye position. **No crash, no log error** — just permanently no head tracking, which on Lume Pad is the entire point of the platform.

Pre-this-PR the only workaround was `adb shell pm grant ...` documented in Step 0 of the bring-up checklist (#350). That works for emulator + developer-mode device but is not a real day-1 path for hardware.

## How it works

```kotlin
class MainActivity : NativeActivity() {
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        if (checkSelfPermission(Manifest.permission.CAMERA) != PackageManager.PERMISSION_GRANTED) {
            requestPermissions(arrayOf(Manifest.permission.CAMERA), REQUEST_CAMERA)
        }
    }
}
```

`super.onCreate(...)` runs the full `NativeActivity` glue (spawns the `android_main` thread, sets up the `ANativeActivity` callbacks, attaches the window) **before** we kick off `requestPermissions`. The native init / `xrCreateInstance` chain runs in parallel with the user-facing dialog. By the time CNSDK reaches its `leia_core_enable_face_tracking` call at session start, the user has already decided.

## Trade-offs

- **First-launch race:** if the user is slow to tap Allow, CNSDK may try to open the camera before the grant lands. Same end-state as current behavior (no face), and re-launching after the grant is a clean second-light. Avoiding this fully would require blocking the native main thread until the permission decision comes back — more invasive than the value it'd add.
- **No `onRequestPermissionsResult` override:** we don't need one. The native code never asks Java about the permission; it just calls into CNSDK which calls into the system camera. If the permission flips state mid-session, CNSDK's `leia_core_enable_face_tracking` either succeeds or no-ops — both are valid states.
- **Deny case:** silent no-face, identical to no-grant-at-all. Not great UX but acceptable for a POC test app; a production app would need to inform the user.

## Verification (Android-36 emulator)

```
$ adb shell pm uninstall com.displayxr.cube_handle_vk_android
$ adb install -r cube_handle_vk_android-debug.apk
$ adb shell dumpsys package com.displayxr.cube_handle_vk_android | grep CAMERA
      android.permission.CAMERA
$ adb shell am start -n com.displayxr.cube_handle_vk_android/.MainActivity
... GrantPermissionsActivity launched ...
$ adb shell pm grant com.displayxr.cube_handle_vk_android android.permission.CAMERA  # simulates "Allow"
$ adb logcat | grep cube_handle
  xrInitializeLoaderKHR -> XR_SUCCESS
  xrCreateInstance -> XR_SUCCESS
  xrCreateVulkanInstanceKHR -> XR_SUCCESS
  xrCreateVulkanDeviceKHR vk_result=-7   # known emulator wall
```

Same full sentinel chain as pre-wrapper. No regression in `NativeActivity` glue (`android_main`, `APP_CMD_INIT_WINDOW`, `xrInitializeLoaderKHR` all reached). The permission dialog launches as the `GrantPermissionsActivity` intent.

## Files

- `test_apps/cube_handle_vk_android/build.gradle` — Kotlin Android plugin, `kotlin-stdlib` dep, `kotlinOptions { jvmTarget = 17 }`.
- `test_apps/cube_handle_vk_android/src/main/java/.../MainActivity.kt` — the wrapper (~10 lines of actual code, rest is the file's why-this-exists block comment).
- `test_apps/cube_handle_vk_android/src/main/AndroidManifest.xml` — drop `android:hasCode="false"`, swap `android.app.NativeActivity` → `.MainActivity`. `android.app.lib_name` meta-data preserved (the wrapper inherits the native lib load path).
- `docs/getting-started/android-bringup-checklist.md` — Step 0 rewritten: dialog is now the default day-1 path; `adb pm grant` becomes opt-in for non-interactive bring-up.

## Test plan

- [ ] On Lume Pad first launch, system permission dialog appears within ~1s of icon tap.
- [ ] Tapping **Allow** results in `dumpsys package ... | grep CAMERA` showing `granted=true`.
- [ ] Head tracking works at session start (face-driven cube parallax visible).
- [ ] Tapping **Deny** results in no head tracking but no crash (graceful degrade — same as missing grant).

### DIFF
--- a/docs/getting-started/android-bringup-checklist.md
+++ b/docs/getting-started/android-bringup-checklist.md
@@ -39,15 +39,12 @@ If A passes you can stop. B confirms the CNSDK convention assumptions. C is only
 
 The test app's `AndroidManifest.xml` declares `android.permission.CAMERA` (CNSDK front-camera face tracking) and `android.permission.WAKE_LOCK` (keep screen on during XR sessions). On Android 6+ `CAMERA` is a "dangerous" permission that requires either a user dialog OR an `adb pm grant`.
 
-The test app is a thin `NativeActivity` with no Java/Kotlin wrapper to surface the runtime permission dialog. Two options for day-1:
+The test app's `MainActivity` is a thin Kotlin wrapper around `NativeActivity` whose only job is to call `requestPermissions(CAMERA)` from `onCreate`. Day-1 path on Lume Pad: launch the app, tap **Allow** when the system permission dialog appears. The wrapper hands off to `NativeActivity` immediately, so the native init / `xrCreateInstance` runs in parallel — by the time CNSDK's face tracker opens the camera at session start, the grant is in place.
+
+For non-interactive bring-up (CI, scripted soak tests, no human at the device), pre-grant via adb:
 
 ```bash
-# Option 1: pre-grant via adb (works on emulator + device-with-developer-mode)
 adb shell pm grant com.displayxr.cube_handle_vk_android android.permission.CAMERA
-
-# Option 2: tap "Allow" on the system permission dialog the first time
-# face tracking tries to open the camera (CNSDK throws the dialog itself
-# on first leia_core_enable_face_tracking call — confirm at A1).
 ```
 
 **Missing CAMERA on Lume Pad symptom:** session reaches FOCUSED, app renders, but `xrLocateViews` always returns the default centered eye position regardless of head motion. No crash, no log error from CNSDK — just silently no face tracking. Check `adb shell dumpsys package com.displayxr.cube_handle_vk_android | grep CAMERA` to confirm grant.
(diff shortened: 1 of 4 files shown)


dfattal:

PR #371, opened 2026-05-31
fix(sim_display): report native monitor resolution on Windows (#369)

### LINKED ISSUE #369: sim_display DP renders only top-left quadrant in workspace/shell mode (works in test apps; Leia DP works)
## Summary

In **workspace / shell mode**, the **sim_display** display processor (DP) renders content only in the **top-left quadrant** of the panel (black everywhere else). The same sim_display DP renders correctly in ordinary (non-workspace) test apps, and the **Leia SR** DP renders the full panel correctly in the *same* shell session. Investigation points squarely at the **runtime** (the workspace DP-handoff geometry + the sim_display DP), not the shell — the shell only registers chrome swapchains and sets window poses; it has no influence over atlas/tile/DP geometry.

## What was verified

The workspace DP handoff lives in `multi_compositor_render()` → `xrt_display_processor_d3d11_process_atlas()` at `src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp:8421-8508`. Per-view dims come from `resolve_active_view_dims()` (`:1646`), which reads the active rendering mode's runtime-computed `view_width_pixels`/`view_height_pixels`. Those are filled by `u_tiling_compute_mode()` (`src/xrt/auxiliary/util/u_tiling.h:42-53`):

```
view  = display × view_scale
atlas = tiles × view
```

using the **plugin-reported display pixel dims** (`target_instance.c:281-291`, runs for both leia-sr and sim-display).

**Leia path (works), display = 3840×2160:**
- Active mode (2×1, scale 0.5×0.5) → view 1920×1080, content 3840×1080, combined atlas 3840×2160.
- Crop triggers (content_h 1080 ≠ atlas_h 2160) → DP gets a tightly-packed 3840×1080 atlas.
- Confirmed in a live service log: `Multi-comp: created crop texture 3840x1080 (view=1920x1080)`, `leia_dp_d3d11_process_atlas weave: target=3840x2160 vp=(0,0 3840x2160) view=1920x1080`.

**sim_display path (broken), display = 1920×1080 (`SIM_DISPLAY_PIXEL_W/H` defaults):**
- Default 3D mode = Anaglyph (mode 1): `view_scale 0.5×0.5`, `tile 2×1` (`sim_display_device.c:567-575`).
- → view 960×540, combined atlas 1920×1080, crop 1920×540. The geometry/crop math itself is correct.

## Leading hypothesis (needs one repro log to confirm)

Because the crop math is correct for both DPs, the quadrant artifact most likely comes from a **second factor specific to sim_display**: its DP **output mode** is a process-global (`sim_display_get_output_mode()`, toggled by the 1/2/3/4/5 keys) that is **independent of the device's active rendering mode**. The five HLSL shaders (`sim_display_processor_d3d11.cpp:52-171`) each assume a specific tile layout:
- Passthrough/SBS: identity / single-tile
- Anaglyph/Blend/Squeezed-SBS: 2×1
- **Quad: 2×2** (`ps_quad_source`, `:108-131`)

Under workspace mode the runtime's V/0-8 key polling is gated off (#303 / `feedback_no_runtime_input_under_workspace`), so the sim_display output-mode global can **desync** from the active rendering mode the compositor tiled with. If the DP is in **Quad (2×2)** mode while the atlas is tiled **2×1**, the quad shader samples a 2×2 grid from a half-populated atlas → content collapses into the **top-left quadrant**, exactly matching the symptom.

Other candidates (lower confidence): combined-atlas WRITE vs crop READ vertical-stride divergence (the documented "atlas stride invariant" hazard — slot stride = `atlas_width/tile_columns`, three coupled callsites); the `content_w/h == atlas_w/h` direct-pass skip at `:8438` misfiring for sim_display's dims.

## Recommended fix direction (final call left to triage)

1. **(Recommended)** Make the sim_display DP output mode track the active rendering mode in workspace sessions, instead of relying on the (gated-off) key-driven global — i.e. derive the shader from `tile_columns/tile_rows` (already passed to `process_atlas`) rather than a stale global. This is sim_display-local and does not touch the shared handoff or the Leia path.
2. If a geometry bug is confirmed instead, fix it in the workspace handoff/crop — but note `:8504` is **shared with the working Leia path**, so any change must be re-validated against Leia to avoid regression.

## To confirm root cause / unblock the fix

A single **sim_display-as-active-DP workspace repro log** disambiguates between the candidates. Run the shell with sim_display as the active DP (Leia plugin not active) and capture from `%LOCALAPPDATA%\DisplayXR\DisplayXR_displayxr-service.exe.*.log`:
- the `Multi-comp: combined atlas WxH` and `Multi-comp: created crop texture WxH (view=...)` lines,
- the `Created sim display D3D11 processor ... initial mode: <X>` line + active output mode,
- (add a one-shot WARN at `:8504` printing `dp_view`, `content`, `atlas`, `tile_columns/rows`, and `sim_display_get_output_mode()` if not already present).

## Evidence pointers
- `src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp:8421-8508` (workspace DP handoff + crop), `:1646` (`resolve_active_view_dims`), `:8438` (content==atlas skip)
- `src/xrt/drivers/sim_display/sim_display_processor_d3d11.cpp:52-171` (5 mode shaders), `:249-252` (output mode read), `:258-259` ("atlas guaranteed content-sized" contract)
- `src/xrt/drivers/sim_display/sim_display_device.c:126-127` (1920×1080 default), `:557-605` (rendering modes + view_scale)
- `src/xrt/auxiliary/util/u_tiling.h:42-53` (tiling formula); `src/xrt/targets/common/target_instance.c:281-291` (mode compute)

_Shell is not implicated; tracking as runtime-side._

dfattal:
Fixes #369.

## Problem
In workspace/shell mode the **sim_display** DP rendered content only in the **top-left quadrant** of the panel (black elsewhere). Standalone test apps were fine, and the Leia SR DP was fine in the same shell session.

## Root cause (confirmed by live repro)
sim_display hardcodes a **1920×1080** panel on Windows (`SIM_DISPLAY_PIXEL_W/H` defaults; the Windows path in `sim_display_device.c` never queried the monitor, unlike the macOS path which uses `CGDisplayPixelsWide`). The workspace compositor sizes the composited atlas to the plug-in-reported resolution and assumes **atlas == panel** — which holds for real vendor plug-ins like Leia SR (it reports the physical 4K size). On a 3840×2160 panel, the half-size 1920×1080 atlas was placed 1:1 into the top-left quadrant of the 3840×2160 back buffer.

Diagnostic instrumentation at the workspace DP handoff confirmed:
- broken (sim default): `atlas(combined)=1920x1080 … bb=3840x2160` → top-left quadrant
- Leia (works): `atlas(combined)=3840x2160 … bb=3840x2160`

Standalone test apps are unaffected because their window is created at the reported size (atlas == window).

## Fix
On Windows, auto-detect the primary monitor's native resolution via `EnumDisplaySettingsW(ENUM_CURRENT_SETTINGS)` (returns true pixels regardless of this process's DPI-awareness). Explicit `SIM_DISPLAY_PIXEL_W/H` env overrides still take precedence. This mirrors the existing macOS auto-detect and satisfies the atlas==panel assumption.

## Verification
On a 3840×2160 Leia panel with sim_display forced as the active DP:
- Before: `atlas=1920x1080`, content in top-left quadrant (user-confirmed screenshot).
- After: sim_display reports/atlases `3840x2160`, content fills the display (verified via `SIM_DISPLAY_PIXEL=3840x2160` override — user-confirmed — and via the auto-detect path: `XR_EXT_display_info (iface=sim-display): pixels=3840x2160, atlas=3840x2160`).

Scope: Windows sim_display only; no change to real vendor plug-ins or the shared DP-handoff path.

### DIFF
--- a/src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp
+++ b/src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp
@@ -11225,9 +11225,17 @@ comp_d3d11_service_create_system(struct xrt_device *xdev,
 		sys->base.info.display_pixel_height = sys->output_height;
 	}
 
-	U_LOG_W("D3D11 service system compositor created: view=%ux%u, view_count=%u, "
-	        "display=%ux%u, output=%ux%u @ %.0fHz",
-	        sys->view_width, sys->view_height, view_count,
+	// These dims are the pre-DP placeholder defaults (set at struct init,
+	// search "output_width = 1920"). The real panel resolution is reported
+	// later by the active display processor — see the "XR_EXT_display_info"
+	// line (display_pixel_*) and the combined-atlas size, which are the
+	// authoritative values. Labelled here so this early line isn't mistaken
+	// for the actual display resolution.
+	U_LOG_W("D3D11 service system compositor created: view_count=%u, "
+	        "default/placeholder dims (pending display-processor query): "
+	        "view=%ux%u display=%ux%u output=%ux%u @ %.0fHz",
+	        view_count,
+	        sys->view_width, sys->view_height,
 	        sys->display_width, sys->display_height,
 	        sys->output_width, sys->output_height, sys->refresh_rate);
 
--- a/src/xrt/drivers/sim_display/sim_display_device.c
+++ b/src/xrt/drivers/sim_display/sim_display_device.c
@@ -37,6 +37,11 @@
 #include <CoreGraphics/CoreGraphics.h>
 #endif
 
+#ifdef XRT_OS_WINDOWS
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#endif
+
 
 /*
  *
@@ -480,6 +485,32 @@ sim_display_hmd_create(void)
 		if (vis_w > 0 && vis_h > 0) {
 			pixel_w = (int)vis_w;
 			pixel_h = (int)vis_h;
+		}
+	}
+#endif
+
+#ifdef XRT_OS_WINDOWS
+	// Auto-detect the primary monitor's NATIVE pixel resolution so the
+	// simulated panel matches the real display the workspace window is
+	// created on. The compositor sizes the atlas to this reported
+	// resolution but assumes atlas == panel (true for real vendor plug-ins
+	// like Leia SR, which report the physical 4K size). Without this,
+	// sim_display's hardcoded 1920x1080 default makes the runtime compose a
+	// half-size atlas onto the full panel, leaving workspace content in the
+	// top-left quadrant (#369). EnumDisplaySettingsW returns true pixels
+	// regardless of this process's DPI-awareness. Explicit env overrides
+	// (SIM_DISPLAY_PIXEL_W/H) still win.
+	{
+		DEVMODEW dm = {0};
+		dm.dmSize = sizeof(dm);
+		if (EnumDisplaySettingsW(NULL, ENUM_CURRENT_SETTINGS, &dm) && dm.dmPelsWidth > 0 &&
+		    dm.dmPelsHeight > 0) {
+			if (getenv("SIM_DISPLAY_PIXEL_W") == NULL) {
+				pixel_w = (int)dm.dmPelsWidth;
+			}
+			if (getenv("SIM_DISPLAY_PIXEL_H") == NULL) {
+				pixel_h = (int)dm.dmPelsHeight;
+			}
 		}
 	}
 #endif
--- a/src/xrt/ipc/server/ipc_server_handler.c
+++ b/src/xrt/ipc/server/ipc_server_handler.c
@@ -127,6 +127,42 @@ validate_device_id(volatile struct ipc_client_state *ics, int64_t device_id, str
 
 #if defined(XRT_HAVE_D3D11_SERVICE_COMPOSITOR)
 /*!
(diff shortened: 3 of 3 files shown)


dfattal:
Two more commits added to this PR (discovered while validating the resolution fix in sim_display workspace mode; will land as separate commits on rebase-merge):

- **`fix(ipc): fill surplus xrLocateViews slots in the SR view-pose path`** — `ipc_try_get_sr_view_poses` only filled the active ≤2 view poses, but `xrLocateViews` reports the system max view count (4 for sim_display's Quad mode). The surplus slots were uninitialized stack garbage → invalid quaternion → `xrEndFrame` rejected every frame → workspace apps stuck on an infinite spinner. Now fills views `[eye_count, view_count)` with valid poses, mirroring the in-process `oxr_session_locate_views` extension. (General correctness fix, not sim_display-specific.)
- **`chore(d3d11_service): label create-time dims as pre-DP placeholders`** — the "service system compositor created" log printed hardcoded 1920×1080 placeholders before the DP reports the real panel size; reworded so it isn't mistaken for the actual resolution.

Both verified live: cube_handle_d3d11 now loads under the workspace in sim_display mode and the display fills at 3840×2160.

Follow-up filed: #372 (workspace chrome actions don't actuate — separate concern, needs runtime-vs-shell triage).

Issue #370, opened 2026-05-31
Shell hit-test Phase 2: move cursor per-eye depth ownership off the runtime raycast


dfattal:
## Context

`displayxr-shell-pvt` #40 Phase 1 ported the eye→cursor ray-vs-window hit-test into the shell for **click policy** (focus / drag / chrome dispatch / resize) — shipped in `displayxr-shell-pvt` PR #50, validated on a real Leia SR display (0/0 divergences vs the runtime's enrichment). That was self-contained controller-side work with no runtime change.

This issue tracks **Phase 2 — the runtime-side change** required to delete `workspace_raycast_hit_test` entirely.

## Why the runtime still raycasts

The runtime renders the mouse cursor sprite into the stereo compositor atlas and needs `hit_z_m` — the ray-plane intersection depth on the window under the cursor — to compute the cursor's **per-eye disparity** so it floats at the same depth as what it's pointing at.

Reference (paths/lines as of writing):
- `src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp`
  - `workspace_raycast_hit_test` (~L5795) — the ~340-line raycast
  - per-frame hover cache + `mc->cursor_hit_z_m` publish (~L6407–6478)
  - cursor sprite draw with per-eye disparity off `hit_z_m` (~L8205–8413)

The shell already computes `hit_z_m` locally (Phase 1) but has no way to feed it to the runtime cursor renderer.

## Options

1. **(smaller)** New runtime path letting the controller push cursor screen-pos + `hit_z` each frame; the existing disparity math runs on the shell-supplied depth. The shell becomes the sole hit-test owner.
2. **(larger)** Controller submits a pre-warped per-eye stereo cursor overlay via the existing cursor swapchain (`xrCreateWorkspaceCursorSwapchainEXT` / `xrSetWorkspaceCursorEXT`, spec_version 13).

Once cursor depth is shell-driven, `workspace_raycast_hit_test` and its non-cursor call sites (POINTER/MOTION event enrichment, the per-frame hover cache that drives POINTER_HOVER, and the public `xrWorkspaceHitTestEXT` API) can be removed.

## Also note

POINTER_HOVER events are themselves generated by the runtime raycast (fired on slot/region transitions). Deleting the raycast means the shell must generate its own hover transitions — which it can only do once it has a per-frame cursor-position feed (Phase 2 gives it one). So hover ownership moves with cursor-depth ownership.

## Acceptance

[...]

PR #379, opened 2026-05-31
feat(workspace): retire residual runtime-side input policy (#376)

### LINKED ISSUE #376: Retire residual runtime-side input policy in workspace mode (DELETE-close, Ctrl+O launch → shell, cursor dim factor)
## Context

Follow-up to #370 (cursor-depth + hit-test moved to the controller, spec_version 22). With hit-test, hover, and cursor depth now fully controller-owned, an audit of the runtime↔workspace contract (`docs/architecture/separation-of-concerns.md`) surfaced a few **residual input/look-and-feel policies still in the runtime** that predate #370 (leftovers from the #305–#308 / ADR-018 migration). They live in `multi_compositor_render` in `src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp` (workspace-mode-only path). Per the contract, workspace apps own all input and the controller owns window behavior + look-and-feel; these belong in the controller.

Cross-repo: items 1 and 2 also touch the shell (`../displayxr-shell-pvt`). Item 3, if done the controller-owned way, is a coupled extension bump (runtime PR first → `publish-extensions` syncs the header to `displayxr-extensions` → re-run + merge the shell PR — the #370 merge order).

## 1. `DELETE` → close focused client
`comp_d3d11_service.cpp` (grep `VK_DELETE` / `request_exit_by_slot`). Window-close is controller policy (ADR-018). The shell's KEY handler does **not** currently handle DELETE, so the runtime is still the de-facto owner of "DELETE = close."
- **Fix (coordinated runtime + shell):** add a `VK_DELETE` case to the shell's `XR_WORKSPACE_INPUT_EVENT_KEY_EXT` handler (`displayxr-shell-pvt` `src/main.c`) → `g_xr->request_client_exit` on the focused client (mirror the existing `request_client_exit` call sites), THEN delete the runtime intercept. Deleting the runtime side *alone* would regress close-on-DELETE.

## 2. `Ctrl+O` app launcher → port to the shell (option b)
This is **not just a keybind** — `comp_d3d11_window_request_app_launch` (`comp_d3d11_window.cpp`) posts `WM_WORKSPACE_LAUNCH_APP`, whose WndProc handler (`comp_d3d11_window.cpp`, ~L984–1045) is a full self-contained legacy launcher:
- `GetOpenFileNameA` "Launch App in Workspace" file browser for any `*.exe`,
- sets `DISPLAYXR_WORKSPACE_SESSION=1` (+ dev `XR_RUNTIME_JSON` under `DISPLAYXR_DEV`),
- `CreateProcessA` spawns it (cwd = app dir, `CREATE_NEW_CONSOLE`).

It is **not** replicated by the shell's Ctrl+L launcher (which launches *curated/registered* apps, not an arbitrary file-browsed exe). **Decision: port the "browse + launch arbitrary exe" capability to the shell** (keep the affordance, move it controller-side).
- **Fix:** in `displayxr-shell-pvt` `src/main.c`, add a `Ctrl+O` case to the KEY handler → `GetOpenFileNameA` (`*.exe` filter) → route the chosen path through the **existing** shell launch path (`shell_launcher_perform` / the same `CreateProcess` + `DISPLAYXR_WORKSPACE_SESSION` + `XR_RUNTIME_JSON`-forwarding the shell already does). The shell already has the spawn infrastructure + `shell_file_picker`.
- Then **delete the runtime side**: the `Ctrl+O` intercept (`comp_d3d11_service.cpp`), `comp_d3d11_window_request_app_launch` + its declaration (`comp_d3d11_window.{cpp,h}`), and the `WM_WORKSPACE_LAUNCH_APP` WndProc handler + message id (`comp_d3d11_window.cpp`) — incl. the now-unused `<commdlg.h>` include if nothing else needs it.

## 3. Hardcoded over-window cursor dim alpha `0.30`
Cursor sprite render block (grep `over_window` / `body_tint`). Minor look-and-feel the controller arguably should own — it already pushes the `overWindow` flag via `xrSetWorkspaceCursorDepthEXT`; it could also push the dim factor.
- **Fix (optional/minor):** add a dim-factor field to `XrWorkspaceCursorDepthEXT` (→ spec_version 22→23 bump + oxr/ipc/service plumbing, mirroring #370), or leave as a documented runtime default. Lowest priority; defer unless wanted.

## Bonus cleanup
- Remove the vestigial dead `mc->drag` field (never set active).
- The drag/resize/rotation state machines ADR-018 flagged as the real violation are already gone (controller-owned via `xrSetWorkspaceClientWindowPoseEXT`).

## Already addressed (not part of remaining scope)
- **Bundled `vulkan-1.dll` in `_package/bin`** — fixed on `main` in `0a28a547f`: vcpkg's applocal-deps install re-staged the runtime/service's `/DELAYLOAD` `vulkan-1.dll`; closed with a guaranteed-last post-install `file(REMOVE)` in `targets/CMakeLists.txt`. (vcpkg applocal has no per-DLL exclude, so a "don't stage it at the source" fix would mean disabling global applocal and hand-rolling filtered installs for every target — disproportionate; the post-install removal is the idiomatic fix and covers all targets.)

dfattal:
Closes #376.

Follow-up to #370. Removes input / look-and-feel policy that lingered in the workspace-mode compositor path (`multi_compositor_render` / the d3d11 window WndProc), per the runtime↔workspace contract — the controller owns input + window behavior + look-and-feel; the runtime is plumbing.

### Changes
1. **DELETE close-focused-client intercept deleted** (`comp_d3d11_service.cpp`). The public `xrRequestWorkspaceClientExitEXT` helper (`comp_d3d11_service_workspace_request_exit_by_slot`) **stays** — it's still the IPC / public-API close path. Only the keyboard intercept is gone; the controller now drives DELETE.
2. **Ctrl+O browse+launch deleted** — the intercept, `comp_d3d11_window_request_app_launch` (+ decl), the `WM_WORKSPACE_LAUNCH_APP` WndProc handler + message id, and the now-unused `<commdlg.h>` include. The browse+launch affordance moves to the controller.
3. **Over-window cursor dim is controller-pushed** — `XrWorkspaceCursorDepthEXT` gains a `dimFactor` field (**spec_version 22 → 23**), threaded through `proto.json`, oxr (clamped to `[0,1]`), the IPC client/server, and the d3d11 service. The cursor render block reads `mc->cursor_dim_factor` (defaulted to the previous hardcoded `0.30` until the first controller push) instead of a literal.
4. **Cleanup** — removes the vestigial dead `mc->drag` field (never set active; drag/resize/rotation are already controller-owned via `set_pose`).

### Coupled change / merge order
Shell PR (`displayxr-shell-pvt`) handles DELETE, ports Ctrl+O, and pushes `dimFactor`. **Merge this runtime PR first** so `publish-extensions` syncs the spec-23 `XR_EXT_spatial_workspace.h` to `displayxr-extensions`; then re-run + merge the shell PR (the #370 merge order).

### Validation
Built locally (`build_windows.bat`, IPC regenerated from proto.json). Deployed into `Program Files\DisplayXR\Runtime` and exercised live with two `cube_handle_d3d11_win` instances against the dev shell: DELETE closes the focused window exactly once; Ctrl+O opens the controller's `*.exe` picker and launches; over-window cursor dim unchanged (`dimFactor = 0.30`). No regressions in Ctrl+L / V / TAB / F11 / ESC / `[` `]`.


dfattal:
Added two follow-up doc commits (repo lean-up requested during review):

- **Prune superseded shell roadmap snapshots** — removed `docs/roadmap/shell-phase1-status.md`, `shell-phase2-plan.md`, and `shell-tasks.md` (completed Phase 0–8 build-time trackers; architecture/design content lives in the active docs + git history). Fixed the dangling links in `CLAUDE.md`, `docs/README.md`, `docs/roadmap/overview.md`.
- **Refreshed ADR-014** — the decision (workspace owns rendering mode) stands; rewrote the stale mechanism (server-side qwerty handler) to the current controller-owned input model + dated Status note.

These are doc-only and adjacent to the #376 review; bundled here to land atomically with the code rather than racing a direct-to-main push while this PR is open.

PR #416, opened 2026-06-04
android(test app): remove dead record_clear()

leaiss:
Static cleanup surfaced during the Android POC pre-hardware pass (deferred item #4).

`record_clear()` was the original `vkCmdClearColorImage`-only per-frame path in `cube_handle_vk_android`. `record_draw()` (the triangle render pass with `LOAD_OP_CLEAR`) fully superseded it and `record_clear()` has had **zero callers** since — confirmed by grep (only the definition + two stale comment references remained).

Drops the function and folds its surviving comment context into `record_draw()`'s header. No behaviour change — never invoked.

**Validation:** `:test_apps:cube_handle_vk_android:assembleDebug` builds clean (arm64-v8a NDK 26.3, BUILD SUCCESSFUL).

### DIFF
--- a/test_apps/cube_handle_vk_android/src/main/cpp/main.cpp
+++ b/test_apps/cube_handle_vk_android/src/main/cpp/main.cpp
@@ -984,91 +984,10 @@ destroy_render_pass()
 
 // ─── per-frame draw ───────────────────────────────────────────────────
 //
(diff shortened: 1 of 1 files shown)


dfattal:
Superseded by #498: cube_handle_vk_android was rewritten (tiled-atlas multiview + crate scene); record_clear() no longer exists. Closing.

PR #417, opened 2026-06-04
android(test app): use ALooper_pollOnce, not deprecated pollAll

leaiss:
Static cleanup from the Android POC pre-hardware pass (deferred item #4). **Stacked on #416.**

`ALooper_pollAll` is deprecated as of **NDK r26** — its internal callback dispatch can silently consume the event the caller was waiting on and return `ALOOPER_POLL_CALLBACK` unexpectedly. `ALooper_pollOnce` has an identical signature; the existing `while(... >= 0)` loop in `android_main` already drains every ready source per iteration, so this is a drop-in that mirrors `android_native_app_glue`'s own migration off `pollAll`.

**Validation:** `:test_apps:cube_handle_vk_android:assembleDebug` builds clean (arm64-v8a NDK 26.3). The poll loop runs pre-XR-init so it executes on the emulator smoke-test path.

### DIFF
--- a/test_apps/cube_handle_vk_android/src/main/cpp/main.cpp
+++ b/test_apps/cube_handle_vk_android/src/main/cpp/main.cpp
@@ -1384,10 +1384,15 @@ android_main(struct android_app *app)
 		// session isn't actively rendering yet — once we hit
 		// SYNCHRONIZED+, we want to spin the frame loop and only
 		// non-blocking poll Android.
+		//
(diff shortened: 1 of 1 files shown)


dfattal:
Closing as part of the post-#498 Android PR cleanup. Note: this change is INDEPENDENT and still applies — main's cube loop still uses the deprecated ALooper_pollAll. It's a trivial one-liner to re-apply on current main if desired.

Issue #263, opened 2026-05-22
Extract drv_leia + Leia plug-in source to its own repo (ADR-019 follow-up)


dfattal:
## Background

Issue #256 / ADR-019 moved Leia SR off the runtime DLL's link line — `dumpbin /imports DisplayXRClient.dll` is now clean. The runtime's *binary surface* is vendor-isolated.

But the runtime's *source tree* still contains the Leia plug-in code:

- `src/xrt/drivers/leia/` — drv_leia + leia_plugin.c
- `installer/DisplayXRLeiaSRInstaller.nsi` — Leia plug-in installer
- `scripts/build_windows.bat` + `.github/workflows/build-windows.yml` — unconditional `gh release download sr-sdk-v$TAG`
- `sr-sdk-v*` GitHub releases hosted under this repo
- Various `XRT_HAVE_LEIA_SR` references in CMake (most are correctly gated, but they only exist because the source lives here)

The plug-in branch lets the runtime *build without* the SR SDK (`-DXRT_HAVE_LEIA_SR=OFF` works — drv_leia + Leia plug-in just don't get built), but CI doesn't exercise that path because the Leia source is co-located. End state per ADR-019 §6: zero vendor source in this repo, zero vendor SDK fetches in this repo's CI.

## Why this matters

With dozens of vendors, hosting every vendor's plug-in source + SDK fetch in the runtime repo is unscalable — organizationally and license-wise (vendor SDKs come with redistribution terms that don't generalize). The runtime should be the OpenXR + IPC + compositor core; each vendor's plug-in should be its own repo, with its own CI, its own vendor-SDK pull, and its own installer release.

## Target shape

```
DisplayXR/displayxr-runtime          ← this repo. Builds with ZERO vendor SDKs.
                                       Ships runtime + sim-display plug-in.

DisplayXR/displayxr-leia-plugin      ← NEW repo. Pulls SR SDK in its own CI.
                                       Ships DisplayXRLeiaSRSetup-*.exe.

DisplayXR/displayxr-vendor-X-plugin  ← future template for new vendors.
```

### Move out of this repo:
- [ ] `src/xrt/drivers/leia/` (drv_leia source + leia_plugin.c + leia_display_processor* + EDID probe + DP factories)
- [ ] `installer/DisplayXRLeiaSRInstaller.nsi`
- [ ] `installer/CMakeLists.txt` Leia plug-in installer wiring (SR_VK_BETA_DLL handling)
[...]

PR #545, opened 2026-06-11
fix(android): source null-compositor frame pacing from panel refresh (was capped at 20 FPS)

leaiss:
## Problem
On Android the OOP runtime uses the **null compositor** (the DP plug-in owns present/weave via the Android surface — there's no swapchain compositor to source vblank from). `target_instance` passed `refresh_rate_hz = 0.0f`, which the null compositor maps to a **20 FPS placeholder default**. That capped `xrWaitFrame` pacing to 20 Hz on every Android OpenXR app, regardless of the 60–144 Hz panel — while the actual per-frame app work is ~10 ms.

This default was a known stopgap: when the in-proc `leia_edid` refresh query was removed in #263, nobody re-sourced the real rate (the old code comment literally said "Add an entry to `xrt_plugin_display_info` if this becomes load-bearing").

## Change
- Add `refresh_mhz` (milli-Hz) to `xrt_plugin_display_info` — append-only, `struct_size`-guarded, ABI-safe.
- `target_instance` now reads `refresh_mhz` from the active plug-in's `get_display_info` **before** creating the null compositor (its pacer is seeded once at create) and uses it to drive `xrWaitFrame` pacing. Falls back to **60 Hz** (not the 20 FPS placeholder) when the plug-in reports nothing.

Pairs with `displayxr-leia-plugin` reporting `refresh_mhz` (that PR depends on this header change).

## Verified
nubia NP02J, OOP runtime: log shows `Null-compositor frame pacing: 60.00 Hz`, `predictedDisplayPeriod` 16.67 ms, and frame rate **20 → ~32 fps** (the new ceiling is the DP weave/present, not the pacer). No regressions to other compositor paths (change is gated to the null-compositor branch).

### DIFF
--- a/src/xrt/include/xrt/xrt_plugin.h
+++ b/src/xrt/include/xrt/xrt_plugin.h
@@ -117,6 +117,15 @@ struct xrt_plugin_display_info
 	/*! Default eye-tracking mode for sessions that don't override.
 	 *  0 = MANAGED, 1 = MANUAL. */
 	uint32_t default_eye_tracking_mode;
+
+	/*! Panel refresh rate in milli-Hz (e.g. 60000 = 60 Hz). Integer so no
+	 *  float crosses the ABI. 0 = unknown → the runtime keeps its own
+	 *  default (the null compositor's frame pacer falls back to 20 FPS).
+	 *  Drives `xrWaitFrame` pacing on the null-compositor (Android OOP)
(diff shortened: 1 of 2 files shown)


dfattal:
Closing as stale ahead of the current release train — this branch is hundreds of commits behind main and hasn't been updated in a while. Not abandoning the idea; please reopen or re-file with a fresh rebase if still wanted.

PR #586, opened 2026-06-15
build(windows): register_dev_plugin.bat — make from-source runtimes find a display processor

dfattal:
## Problem
Windows plug-in discovery is **registry-only** (`HKLM\Software\DisplayXR\DisplayProcessors`) — no adjacent-dir / `XRT_PLUGIN_SEARCH_PATH` fallback, unlike POSIX (`target_plugin_loader.c`). `build_windows.bat` builds the sim-display plug-in DLL but never registers it, so a **pure from-source runtime finds no display processor** → `xrt_instance_create_system` fails (`XRT_ERROR_DEVICE_CREATION_FAILED`, or the app reports "Failed to initialize OpenXR"). A contributor doing a from-source build (no installer) was stranded on exactly this.

Also: a from-source runtime that's *ahead of the released ABI* will ABI-reject an installed (released) vendor plug-in, so "install the bundle first" isn't a reliable substitute.

## Fix (stopgap)
`scripts\register_dev_plugin.bat` registers the freshly-built, ABI-matched plug-in exactly as CI (`build-windows.yml`) and the installer do:
- default → dev **sim-display** (ProbeOrder 200, fallback, no hardware)
- `leia <dll>` → dev-built **Leia** plug-in (ProbeOrder 50)
- `list` → `displayxr-cli dp list`

Plus a CLAUDE.md note so the next source-builder isn't stuck. Elevated prompt required (HKLM).

## Not in scope (tracked separately)
The deeper asymmetry — Windows has no adjacent-dir fallback, so sim-display isn't an automatic universal fallback the way it is on POSIX — is a design decision (registry-only is arguably intentional for control/security). Tracked in a follow-up issue.

Windows-only `.bat`; authored on macOS, **needs a Windows smoke test** (CI doesn't exercise it). @cyrusrohani this is your immediate unblock.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -202,6 +202,14 @@ _package\run_cube_handle_d3d11_win.bat
 
 After a rebuild, copy runtime binaries into `C:\Program Files\DisplayXR\Runtime` (registry discovery finds it); only run the installer when the installer itself changed.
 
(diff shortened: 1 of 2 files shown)


dfattal:
Folded into #588 (one-line dev-setup umbrella) — register_dev_plugin.bat is included there and is what dev-setup.bat calls. Closing in favor of #588.

Issue #396, opened 2026-06-02
Epic: shared helper module (displayxr-common) + unified capture API — kill the 6× Kooima/common drift


dfattal:
﻿## Why this epic

The same native helper code — above all the **Kooima off-axis projection math** (`display3d_view.{c,h}`, `camera3d_view.{c,h}`, `leia_math.h`) — is **vendored by copy in at least six places** across the org and is already drifting (the clip-plane fix in #393 had to be hand-applied to one copy with no way to sync the others). The #389 work (VK native compositor now reliably composites many window-space layers; new `windowspace_handle_*_win` test apps) turns the window-space-layer + HUD + input scaffolding into a genuinely reusable building block too. This epic makes the shared code a **versioned dependency** so editing-a-local-copy becomes structurally impossible.

#393 originally scoped this to the two C++ demos and excluded the engines. That was too narrow: the engine plugins compile the **same math C files**. This epic supersedes that framing; #393 is re-scoped to be layer 2 (the C++ scaffolding) of the design below.

## Drift inventory — every current copy

| Copy location | Repo | Form | Notes |
|---|---|---|---|
| `test_apps/common/` | displayxr-runtime | C/C++ | this repo's native test apps |
| `common/` | displayxr-demo-modelviewer | C/C++ | Vulkan `[0,1]` projection fix landed here |
| `common/` | displayxr-demo-gaussiansplat | C/C++ | + unmerged `feat/zdp-clip-soft-fade-pick` WIP (near/far, selftest, center_view) |
| `include/` | **kooima-projection** | C | the intended math home — seeded then abandoned (last push 2026-05-04) |
| `Source/DisplayXRCore/Private/Native/` | **displayxr-unreal** | C | compiled into the UE module (see ADR-003 UE-native off-axis projection) |
| `native~/` (`display3d_view.h`, `displayxr_kooima.{h,cpp}`) | **displayxr-unity** | C++ → P/Invoke | wraps the same Kooima math (see ADR-006 window-relative Kooima) |

~20 of the ~28 `common/` files are byte-identical between the two demos; nearly all divergence is accidental.

## Architecture — one repo, two CMake targets (no new repo)

Repurpose the dormant **`kooima-projection`** repo → rename to **`displayxr-common`**, exposing two targets so the math core is shared by *everything* while engines never link the C++/Win32 scaffolding:

```
displayxr-common   (renamed from kooima-projection — net new repos: 0)
[...]

PR #591, opened 2026-06-15
fix(#396): qwerty/legacy P-toggle is pure disturbance-free (camera-centric default)

dfattal:
## What

Makes the legacy/qwerty **P** toggle a pure disturbance-free rig switch — the same behavior as the test-app **C** toggle, but starting camera-centric.

- **Default is already camera-centric** with the known defaults (`camera_mode=true`, 0.5 dp convergence, 0.3249 half-tan vFOV, m2v 1; display side vHeight 1.3). Unchanged.
- **P used to** do the seamless converter switch on frame 0 **then ease the params to the mode defaults over 0.4 s** — that easing was a visible change.
- **P now** just switches the rig and holds the view. `display↔camera↔display` is an identity (with the v1.0.4 converter fix on main), so pressing P repeatedly never moves the scene. Per-mode defaults still apply at initial startup, not on toggle.

`qwerty_advance_rig_anim` is now inert (left in place, harmless).

## Pin / dependencies

**Runtime stays on common v1.0.4** — it links `displayxr::math_xrt` (the converter wrapper), not the test-app `displayxr::common`. v1.0.5's additions (rig_mode/ViewParams/input_handler) are test-app-only; the `dxr_view_math` converter math is unchanged. No pin bump here. (Depends on the v1.0.4 converter correctness fix already on `main`.)

## Validation

MinGW `drv_qwerty` compile-check passes.

## 🪟 Windows test (merge when green)
1. `scripts\build_windows.bat build` then run a **legacy/hosted** app (e.g. `_package\run_cube_hosted_d3d11_win.bat`), or any app under the qwerty path.
2. Confirm it **starts camera-centric** with the expected look (0.5 dp / 36° vFOV).
3. Press **P** repeatedly → rig label flips Camera↔Display but the **view does not move/zoom/ease**. (Previously it eased to defaults over ~0.4 s.)
4. Adjust convergence/zoom/vHeight, then **P** → still seamless from the current look (no snap to defaults).

Merge (rebase) when green.

### DIFF
--- a/src/xrt/auxiliary/math/CMakeLists.txt
+++ b/src/xrt/auxiliary/math/CMakeLists.txt
@@ -20,7 +20,7 @@ set(DISPLAYXR_XRT_INCLUDE_DIR "${_dxr_xrt_includes}" CACHE PATH "" FORCE)
 FetchContent_Declare(
 	displayxr_common
 	GIT_REPOSITORY https://github.com/DisplayXR/displayxr-common.git
-	GIT_TAG v1.0.4
+	GIT_TAG v1.0.7
 	GIT_SHALLOW TRUE
 	)
 FetchContent_MakeAvailable(displayxr_common)
--- a/src/xrt/drivers/qwerty/qwerty_device.c
+++ b/src/xrt/drivers/qwerty/qwerty_device.c
@@ -654,6 +654,7 @@ qwerty_system_create(struct qwerty_hmd *qhmd,
 	qs->disp_spread_factor = 1.0f;
 	qs->disp_parallax_factor = 1.0f;
 	qs->disp_vHeight = 1.3f; // 1.3m
+	qs->disp_perspective = 1.0f;
 
 	// Hardware config defaults (overridden by target builder)
 	qs->nominal_viewer_z = 0.6f;  // meters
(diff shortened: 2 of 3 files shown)


dfattal:
## Fix pushed — perspective was being dropped on the toggle

Repro confirmed: P went camera→display but landed on the display rig's **own** perspective=1 default instead of the camera-equivalent — a visible jump.

**Root cause:** qwerty hardcoded the display rig's `perspective_factor = 1.0` (in `qwerty_get_view_state` and as the display→camera input) and never stored the camera→display converter's `perspective_factor` output. The camera default (0.5 dp / 36° vFOV) converts to perspective ≈ **0.50** at the legacy panel/distance (H=0.194, N=0.6), with vHeight ≈ 1.30 — not perspective 1.0. The old animate-to-defaults masked it; removing the animation exposed the incomplete conversion.

**Fix:** new `qs->disp_perspective` (default 1.0) — store the converted perspective on camera→display, feed it back as the display→camera input, report it from `qwerty_get_view_state`.

**Offline proof** (qwerty defaults): converted display rig `vHeight=1.30, perspective=0.498` → display view matches camera view **Δ=0** (fov + eye_world); with the old fixed perspective=1.0 the same eye is off by **−8.8° vFOV and 2.0 m depth** — exactly the jump. MinGW `drv_qwerty` passes.

Please re-pull and re-test: start (camera, 0.5 dp look) → **P** → display rig should look identical (no jump), and **P** again → back to camera identically.

Issue #510, opened 2026-06-09
Android: out-of-process (service) runtime so apps are CNSDK-free — vendor-isolation parity with Windows


dfattal:
## Problem

On Android, running an in-process OpenXR app against the **Leia** DP currently forces the **app** to package the CNSDK glue (the `sdk-<ver>.aar` — via `cnsdk.dir`/`-PcnsdkDir`). Without it the app JNI-aborts with `NameNotFoundException: com.leialoft.display.config` the moment CNSDK tries to reach the Leia system services.

This breaks the intended boundary — **the app should talk only to DisplayXR (OpenXR), never to a vendor SDK** — which already holds on Windows.

## Root cause

The cube uses the **`in_process`** runtime flavor (`…openxr_runtime.in_process`), so the whole stack dlopens into the **app's own process**:

```
app → openxr_displayxr.so (runtime) → libdxrp050_leia_cnsdk.so (Leia DP) → libleiaSDK-jni.so (CNSDK)
```

CNSDK must reach the Leia system-service APKs (`com.leialoft.display.config`, `com.leia.headtrackingservice`). Under **Android 11+ package visibility**, the *calling process's manifest* must declare `<queries>` for those packages — and since CNSDK runs in the app's process, it's the **app's** manifest that must carry them. The CNSDK aar injects exactly that (`<queries>` + JNI Java glue). Hence CNSDK leaks into the app.

Windows has no package-visibility model and the SR weaving runs in the runtime's module (COM/named-pipes), so SR/CNSDK never loads into the app's address space — the app stays purely OpenXR.

## Proposed fix

Make the **`out_of_process`** flavor (`…openxr_runtime.out_of_process`) the supported path for vendor DPs on Android:

- Runtime + Leia plugin + CNSDK live in the DisplayXR runtime **service** process; the app connects over IPC and hands its surface across (the surface-injection path already exists: `MonadoImpl.nativeAppSurface` in `src/xrt/targets/service-lib/service_target.cpp`).
- The Leia `<queries>` move to the **runtime service's** `AndroidManifest.xml` (`src/xrt/targets/openxr_android/src/main/AndroidManifest.xml`), where ADR-019 vendor-isolation says they belong — **not** in the app.
- Net: the app's process never loads CNSDK → app is purely OpenXR, parity with Windows.

## Scope / acceptance

[...]

PR #646, opened 2026-06-24
fix(android,#510): closed-loop compositor pacing via AChoreographer vsync

leaiss:
## What

Closed-loop compositor frame pacing on Android (OOP), fixing a cold-start
**30 Hz lock** on the gauss-splat / per-session present path (#510).

## Why

Measured at the display with `dumpsys SurfaceFlinger --latency`: the compositor
presented at a **steady 33.4 ms on a 60 Hz panel** — every 2nd vblank, a hard
30 Hz lock. Cause: the **fake compositor pacer is open-loop on Android**. It
re-syncs its present-time phase grid from real vblank only via
`VK_EXT_display_control` (direct-mode only) or `VK_GOOGLE_display_timing`
(absent here) — Android has neither, so the phase is anchored once at cold start
(during the ~1.5 s surface extent-settle flap) and never corrected, landing
just-after each vblank forever. A manual minimize→maximize cures it only because
the surface teardown recreates the pacer with a fresh anchor.

## How

Give the pacer a real vsync signal:
- An **`AChoreographer` thread** (`comp_target_swapchain`, Android-only) feeds
  frame-callback timestamps into `cts->vblank.last_vblank_ns`.
- The OOP per-session path calls `comp_target_update_timings()` each frame so
  `do_update_timings_vblank_thread` → `u_pc_update_vblank_from_display_control`
  feeds the pacer — the **same closed-loop mechanism direct mode uses**.

**Measured result:** present cadence goes from steady 33.4 ms to predominantly
**16.7 ms (60 Hz)** at cold start.

## Scope / risk

- **Android-only.** The Choreographer thread + its start in `create_images`, the
  `DXR_ANDROID_VSYNC_PACING` flag, and the new `comp_target_update_timings()`
  call are all `#ifdef XRT_OS_ANDROID`. Windows/macOS build + behavior unchanged
  (verified: local Windows build clean).
- Behind `DXR_ANDROID_VSYNC_PACING` (default **on**; set `0` to revert to the
  open-loop behavior).
- Pairs with `U_PACING_APP_USE_MIN_FRAME_PERIOD=1`, which separately stops the
  app pacer doubling 16.7→33.3 ms at cold start.

## Known residual (not addressed here)

A periodic cold-start hitch remains when the CPU DVFS governor parks the big
cores at minimum clock (no touch/interaction) — the frame then can't fit in
16.7 ms even with correct phase. Only a fresh client `SurfaceView` (e.g. manual
minimize/maximize) or a sustained-performance hint clears it. Tracked
separately.

### DIFF
--- a/src/xrt/compositor/main/comp_target_swapchain.c
+++ b/src/xrt/compositor/main/comp_target_swapchain.c
@@ -25,6 +25,8 @@
 #include "android/android_globals.h"
 #include "util/u_time.h"
 #include <android/native_window.h>
+#include <android/choreographer.h>
+#include <android/looper.h>
 #include <sys/system_properties.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


leaiss:
@dfattal — review request, high level:

**Problem (#510):** Android OOP apps cold-launch locked at **30 fps**. Measured with `dumpsys SurfaceFlinger --latency`: a steady **33.4 ms cadence on a 60 Hz panel** — the compositor presents on every *2nd* vblank.

**Root cause:** the fake compositor pacer is **open-loop on Android**. It only re-syncs its present-phase grid from real vblank via `VK_EXT_display_control` (direct-mode only) or `VK_GOOGLE_display_timing` — Android has neither. So the phase is anchored once at startup (during the surface extent-settle flap) and never corrected → it lands just-after each vblank forever → FIFO → 30 Hz. (A manual minimize/maximize "fixes" it only by recreating the pacer.)

**Fix:** feed the pacer the vsync signal it's missing. An `AChoreographer` thread pushes real frame timestamps into the existing `u_pc_update_vblank_from_display_control` plumbing — the **same closed-loop path direct mode already uses** — and the OOP per-session path now calls `comp_target_update_timings()` per frame (it previously didn't). **Result: cadence 33.4 ms → predominantly 16.7 ms (60 Hz)** at cold start.

**Why it's safe to approve:**
- **Android-only** — every hunk is `#ifdef XRT_OS_ANDROID`; Windows/macOS build + behavior unchanged (local Windows build clean; CI confirms).
- **Gated + reversible** — `DXR_ANDROID_VSYNC_PACING` (default on; `0` = old open-loop behavior).
- Reuses existing pacing machinery; no new model. 1 commit, +114 lines, 2 files.

**Out of scope (tracked separately):** a residual cold-start hitch from the CPU DVFS governor parking the big cores at min clock with no touch input — orthogonal to pacing phase; needs a fresh surface / perf-hint.

Issue #666, opened 2026-06-30
Code-sign Windows releases to pass Smart App Control (SAC)


dfattal:
## Problem

On Windows 11, **Smart App Control (SAC)** blocks our unsigned binaries: a "blocked" toast appears, SAC flags DisplayXR files as unsigned / unknown-source, and content fails to render. Disabling SAC confirms the root cause. We should code-sign Windows releases so the stack runs with SAC enabled.

## Why this happens

SAC blocks code that is neither signed by a reputable cert **nor** known-good in Microsoft's Intelligent Security Graph (ISG). Critically, **SAC blocks at DLL-load, not just at the installer `.exe`** — so the entire load chain must be signed, not only the installer. Our chain is currently unsigned: the runtime DLLs, the vendor plug-in DLL, the Unity native plugin (`displayxr_unity.dll`), the MCP framework DLLs, and the shell exe.

## What "signing" buys us (nuance)

Signing is **necessary but not automatically sufficient** for SAC — SAC is reputation-based:

- A new **OV cert** can still be blocked until signed binaries accrue ISG reputation.
- An **EV cert** establishes trust fast (often immediately) — the reliable SAC path.
- **Azure Artifact Signing** (MS-run, flat-fee tier, CI-native) is well-trusted by SAC, but org identity validation has eligibility gates → an EV cert from a CA is the fallback.

## Scope — cross-cutting

Signing only the bundle `.exe` will NOT help — SAC still blocks the unsigned DLLs underneath. Per-repo signing coverage:

- [x] `displayxr-runtime` — runtime DLLs + `DisplayXRSetup-*.exe` (SIGN_CMD-gated, branch `feat/release-signing-664`)
- [x] `displayxr-leia-plugin` — plug-in DLL + installer (same plumbing)
- [x] `displayxr-unity` — `displayxr_unity.dll` (in `.tgz` + `upm` branch)
- [x] `displayxr-unreal` — packaged `Binaries/Win64` DLLs
- [ ] `displayxr-mcp` — framework DLLs + installer
- [ ] `displayxr-shell-pvt` → releases — shell exe + installer
- [ ] `displayxr-demo-*` — demo binaries + installers
- [ ] `displayxr-installer` — meta-installer bundle `.exe`

## Approach (implemented for the first four repos)

[...]

PR #669, opened 2026-06-30
feat(signing,#666): SIGN_CMD-gated release signing + two-pass signed uninstaller

dfattal:
SIGN_CMD-gated code signing for the runtime installer, plus two correctness fixes found while validating it end-to-end with a self-signed cert. _(Tracking issue #666, supersedes #664.)_

## What this does
- **`scripts/sign-release.ps1`** (new): signs every unsigned `*.dll`/`*.exe` in a folder via `$SignCmd`, leaves already-validly-signed third-party DLLs alone, re-verifies the tree.
- **`build_windows.bat`**: when `SIGN_CMD` is set, signs `_package\bin` before NSIS, then the installer `.exe` after. Unset → unsigned build, exactly as before.
- **`installer/DisplayXRInstaller.nsi` + `CMakeLists.txt`**: `!finalize` signs the installer; the uninstaller is signed via a **two-pass** build (below). `SIGN_CMD` cache var is `FORCE`d so the live env always wins on reconfigure.
- **`/release` + `/dxr-release` skills**: capability-gated local-signed-build-replaces-CI-asset phase.

No cert or secret lives in the repo. Without `SIGN_CMD` (clone/CI) the build is unsigned and unchanged.

## Fixes (vs the original signing commit)
- **FORCE the SIGN_CMD cache** — a plain `CACHE` var is seeded only on the first configure, so a tree first configured unsigned kept caching `""` and silently shipped an unsigned uninstaller.
- **two-pass signed uninstaller** — `!uninstfinalize` is unreliable: the install-time uninstaller is a self-copy of the signed installer header and inherits a dangling cert-table pointer → effectively unsigned. Replaced with the canonical two-pass build kept in the `.nsi` (`!makensis` builds an INNER installer that emits the uninstaller → `!system` signs it → the real pass File-includes it). CI/unsigned path unchanged.

## Validation
Self-signed cert, no signing dongle: unsigned path unchanged; all `_package` binaries Valid + `signtool verify /pa`; installed `Uninstall.exe` Valid + executes. Also proven in a 3-component à-la-carte signed install (runtime + plug-in + shell) with the display-processor plug-in active on real hardware.

⚠️ CI builds unsigned, so green CI validates the build + the unsigned `WriteUninstaller` fallback — not signing. Signing is validated locally and happens at release.

### DIFF
--- a/.claude/skills/dxr-release/SKILL.md
+++ b/.claude/skills/dxr-release/SKILL.md
@@ -180,8 +180,64 @@ CI_CONC="${S#completed/}"
 ```
 
 ### Step 3.3: Branch on outcome
-- `success` → Phase 4
+- `success` → Phase 3.5
 - else → STOP, report failed jobs via `gh run view "$RUN_ID" -R "$REPO" --log-failed`. No rollback — tags are sticky; user retries with a new tag.
+
+---
+
+## PHASE 3.5: CODE-SIGN THE COMPONENT INSTALLER (capability-gated)
+
+Same model as the runtime's `/release` skill: GitHub-hosted CI builds the
+component **unsigned** (the code-signing key is held on a build machine and
+isn't available to cloud CI runners, and no secret lives in these public
+repos). A signed release is produced by a **local signed build on a
+signing-capable machine** that replaces the CI asset. No-ops cleanly without
+signing capability.
+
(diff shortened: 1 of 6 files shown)


dfattal:
Tracking issue is **#666** (supersedes #664). Commit subjects retain `#664` — kept as-is to avoid force-pushing the branch; canonical issue is #666.

Issue #706, opened 2026-07-05
Linux on-screen HW validation (Phase 1b/2b/3b + demos) — results


dfattal:
Report surface for the Linux **on-screen** validation. Runtime + all 5 demos are build-green on CI; this verifies they actually present/weave on a real Linux + Vulkan-GPU box.

**Protocol (follow this):** [`docs/getting-started/linux-hardware-validation.md`](https://github.com/DisplayXR/displayxr-runtime/blob/main/docs/getting-started/linux-hardware-validation.md)
**One command per test:** `./scripts/run_linux_demo.sh <target>` (see `--list`).

Related: #660 (runtime), #699 (demos), #705 (packaging, blocked on this).

---

### Environment (paste output)
```
# uname -a; lsb_release -d; echo DISPLAY=$DISPLAY XDG_SESSION_TYPE=$XDG_SESSION_TYPE;
# vulkaninfo --summary | grep -E 'deviceName|driverName|apiVersion'
<paste here>
```

### Results
- [ ] **1. cube-hosted** (Phase 1b — runtime-created window) — ✅/❌ + note
- [ ] **2. cube-handle** (Phase 3b — app X11 window via XR_EXT_xlib_window_binding) — ✅/❌ + note
- [ ] **3. service / IPC** (Phase 2b — out-of-process) — ✅/❌ + note
- [ ] **4. mediaplayer** — ✅/❌ + note
- [ ] **5. gaussiansplat** — ✅/❌ + note
- [ ] **6. modelviewer** — ✅/❌ + note
- [ ] **7. avatar** — ✅/❌ + note
- [ ] **8. earthview** — ✅/❌ + note (needs a 3D-Tiles API key; skip+note if unavailable)

For each: attach a screenshot (`scrot`) on pass; paste the `tee` log excerpt on fail. Also note anything from the "Known first-run risks" section (VK-native path, XCB present support, swapchain format/extent, resize).

A failure with a good log + environment is as useful as a pass — it's the next fix.

hshakula:
# Linux hardware validation — results

[...]

Issue #699, opened 2026-07-05
Demos → Linux (meta): port all 5 displayxr-demo-* to Linux


dfattal:
Meta-tracker for porting the standalone demos to Linux, under the **M8: Linux Support** epic. Sibling of the runtime tracker #660.

**Playbook:** [`docs/guides/linux-demo-port.md`](https://github.com/DisplayXR/displayxr-runtime/blob/main/docs/guides/linux-demo-port.md) — follow it for every demo.

### Key finding — all demos are Vulkan
Audited 2026-06: `avatar`, `earthview`, `gaussiansplat`, `mediaplayer`, `modelviewer` all render with Vulkan. So **no GL compositor backend is needed** — every demo rides the Phase 1 `vk_native` + `VK_KHR_xcb_surface` path, and each port is **pure build tooling** (Linux build script + OpenXR-loader provisioning + CMake Linux arm + CI), same recipe as the runtime's `cube_hosted_legacy_vk_linux`.

### Two halves per demo
- **build-green** — compiles on `ubuntu-latest` (no GPU/display). Parallelizable now.
- **on-screen** — presents + weaves against the dev Linux runtime. Gated on runtime **Phase 1b** (on-screen present) + hardware.

### Per-demo issues
- [ ] mediaplayer — DisplayXR/displayxr-demo-mediaplayer#30
- [ ] gaussiansplat — DisplayXR/displayxr-demo-gaussiansplat#60
- [ ] modelviewer — DisplayXR/displayxr-demo-modelviewer#40
- [ ] avatar — DisplayXR/displayxr-demo-avatar#21
- [ ] earthview — DisplayXR/displayxr-demo-earthview#19

### Sequencing
Do all five **build-green** ports now (parallelizable, CI-validated); run the **on-screen** cascade once runtime Phase 1b lands on hardware. Engines (Unity/Unreal), CEF, and shell are **out of scope** for M8 — deferred to a later Phase 4.

dfattal:
### ⚠️ Correction — demos depend on Phase 3

[...]

Issue #660, opened 2026-06-28
Linux support (tracking): headless substrate → Vulkan/XCB compositor → IPC


dfattal:
## Linux Support (tracking)

Windows, macOS, and Android ship for most features; Linux is the remaining platform. Full plan: [`docs/roadmap/linux-support.md`](../blob/main/docs/roadmap/linux-support.md).

**TL;DR:** the non-presentation substrate survived the Monado fork intact and still compiles behind the non-Windows branches (IPC AF_UNIX sockets + SCM_RIGHTS fd-passing, shmem, plug-in discovery, handle typedefs, service mode). The **entire gap is presentation** — no native compositor can put pixels on a Linux display — plus build tooling. So the port is narrow and well-bounded.

### Phase 0 — Configure + headless self-test
Prove the substrate before touching pixels: OS detection, `aux/os`, the POSIX plug-in loader, the sim_display DP, and a valid display-info report — all hardware-free (`displayxr-cli` links `target_instance_no_comp`, so no compositor or OpenXR loader needed).

- [ ] `scripts/build_linux.sh` — build runtime + `displayxr-cli` + sim_display plug-in, stage plug-in + manifest, run `displayxr-cli selftest` *(drafted)*
- [ ] Relax `find_package(udev REQUIRED)` so a box without `libudev-dev` configures *(drafted)*
- [ ] **Run it green on a stock Ubuntu box** (needs Linux hardware — not yet executed)
- [ ] Install rule: drop sim_display `.so` + manifest into XDG / `/usr/share/displayxr/DisplayProcessors` roots
- [ ] `build-linux.yml` CI job running the selftest as a required check

### Phase 1 — In-process presentation via `vk_native` + XCB
XCB before Wayland (X11 exposes absolute window position, which the DP needs for interlacing-phase tracking).

- [ ] Add `LINUX`/`UNIX` to the `vk_native` CMake gate (`compositor/CMakeLists.txt:339`, `vk_native/CMakeLists.txt:8`)
- [ ] `VK_KHR_xcb_surface` branch in the Linux `#else` at `comp_vk_native_target.cpp:887` + resize/swapchain-recreate
- [ ] `comp_vk_native_window_xcb` for the `_hosted` class
- [ ] First runnable target: `cube_hosted_vk_linux` with sim_display weaving

### Phase 2 — Service / IPC path
[...]

Issue #692, opened 2026-07-03
test_apps/: fix orphaned apps, unify build wiring, regroup by class


dfattal:
Part of #691 (2026-07 ecosystem audit).

Reorganize `test_apps/` — three problems, in priority order. **Do #1 immediately** (it's a silent coverage-loss bug), then the wiring unification, then the folder regroup. Do **not** attempt to merge per-API `main.cpp`s in the same pass (behavior-affecting rewrite; later, per-class, if at all).

## 1. Fix silently-unbuilt apps (do now, independent of the reorg)
- `cube_zones_texture_{d3d12,gl,vk}_win` exist on disk and get run-scripts generated (`scripts/build_windows.bat:466,480`) but are **absent from the `add_subdirectory` list** in `test_apps/CMakeLists.txt:34-55` → they never build on the standard Windows path.
- `cube_hosted_legacy_vk_linux` is built by no supported path — delete or explicitly park.

## 2. Unify the dual build wiring
Windows builds from a hardcoded `add_subdirectory` list (`test_apps/CMakeLists.txt:34-55`, guarded `if(NOT WIN32) return()`); macOS builds from explicit per-app `cmake` calls + hand-written `run_*.sh` heredocs in `scripts/build_macos.sh:115-192, 425-562`. These are manually kept in sync and have already drifted (see #1).
- Replace both hardcoded lists with **one platform-aware aggregator** that discovers `*/*/CMakeLists.txt` and gates each app on its dir suffix (`_win`→WIN32, `_macos`→APPLE, `_android`→ANDROID, `_linux`→LINUX). This makes the "orphaned app" failure mode structurally impossible.
- Generate run-scripts for **both** OSes from a single `apps.cmake` table (app → class/api/platform/env). Fixes the macOS mismatch where `run_cube_handle_metal.sh` drops the `_macos` suffix of its app dir.

## 3. Regroup by handoff class (leaf names unchanged)
Class is the right primary axis: the app-classes doc, `/new-displayxr-app`, and every compositor-path decision key on class; API is a renderer detail inside the app.

```
test_apps/
  CMakeLists.txt        # single platform-aware aggregator (globs, gates on suffix)
  apps.cmake            # the ONLY place run-scripts (both OSes) are generated from
  common/               # leia_math.h + textures + displayxr-common shim
  handle/   cube_handle_*        cube_zones_*        # zones = FEATURE modifier, not a class
[...]

PR #708, opened 2026-07-06
Fix Linux build of the native Vulkan compositor + cube test apps

hshakula:
## What

Five minimal fixes that let the runtime and the two Linux cube test apps
actually **build on a real Linux desktop**. Found while doing the on-hardware
Linux validation in #706 (runtime #660 / demos #699).

## Why CI is green but a real desktop isn't

The Linux CI gate runs the headless Phase 0 build, on a box without EGL/GL dev
libs and with Vulkan-Headers that don't carry the portability macros. Under
those conditions the native-VK-compositor translation units below are either
not compiled at all or don't reference the missing macros — so CI stays green
while `build_linux.sh --apps` on an actual Vulkan+X desktop fails hard.

## The fixes

| File | Problem | Fix |
|------|---------|-----|
| `src/xrt/compositor/gl/comp_gl_compositor.cpp` | The GLAD include block (comment: *"cross-platform"*) and the compositor struct's *Platform context* section have `XRT_OS_WINDOWS` / `XRT_OS_ANDROID` / `__APPLE__` branches but **none for desktop Linux** and no `#else`. Result on Linux: every `GL*` type undeclared + `owns_window` undeclared → **653 errors**. | Add `XRT_OS_LINUX` branches taking the EGL path (mirrors the Android branch). |
| `src/xrt/compositor/util/comp_vulkan.c` | `VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME` / `VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR` (a MoltenVK-only opt-in) referenced unconditionally; not defined by Ubuntu's Vulkan-Headers v204, not needed on Linux. | `#ifdef`-guard on the macro so it compiles out where undefined; unchanged on macOS/MoltenVK. |
| `test_apps/legacy/cube_hosted_legacy_vk_linux/main.cpp` | Same portability macros, same problem. | Same `#ifdef` guard. |
| `test_apps/cube_handle_vk_linux/main.cpp` | Same portability macros, same problem. | Same `#ifdef` guard. |
| `scripts/build_linux.sh` | `--apps` looks for `test_apps/cube_hosted_legacy_vk_linux`, but that app lives under `test_apps/legacy/` (only `cube_handle_vk_linux` is top-level). | Fall back to a nested `find` when the flat path is absent; error clearly if truly missing. |

Each source hunk is annotated with a `CANDIDATE PATCH (#706 Linux validation)` comment so maintainers can quickly review the intent. The EGL-path choice for the GL compositor's Linux branch mirrors Android and matches the build's `XRT_HAVE_EGL`/`XRT_HAVE_OPENGL`; if the project intends a GLX path instead, that's an easy swap.

## Verified on

- Ubuntu 22.04.5, Vulkan 1.3.204, NVIDIA RTX 3080 (+ AMD Renoir), X11/XWayland.
- Runtime + both cube apps build clean (`build_linux.sh --apps`), and the **hosted** and **handle** (`XR_EXT_xlib_window_binding`) sessions bring up the native VK compositor and render the stereo textured cube (verified via the compositor's atlas readback). `--service` build also green. Full results in #706.

### DIFF
--- a/scripts/build_linux.sh
+++ b/scripts/build_linux.sh
@@ -190,7 +190,16 @@ if [ "$BUILD_APPS" = "ON" ]; then
   #                                 passes it via XR_EXT_xlib_window_binding.
   for APP in cube_hosted_legacy_vk_linux cube_handle_vk_linux; do
     APP_DIR="$ROOT/test_apps/$APP"
-    echo "=== Building $APP ==="
+    # CANDIDATE PATCH (#706 Linux validation): the apps aren't all flat under
+    # test_apps/ — cube_hosted_legacy_vk_linux lives in test_apps/legacy/. Fall
+    # back to a nested lookup when the flat path is absent.
+    if [ ! -d "$APP_DIR" ]; then
+      APP_DIR="$(find "$ROOT/test_apps" -type d -name "$APP" | head -1)"
+    fi
+    if [ -z "$APP_DIR" ] || [ ! -f "$APP_DIR/CMakeLists.txt" ]; then
+      echo "ERROR: test app source for '$APP' not found under $ROOT/test_apps" >&2; exit 1
+    fi
+    echo "=== Building $APP (src: $APP_DIR) ==="
     cmake -B "$APP_DIR/build" -S "$APP_DIR" -G Ninja \
       -DCMAKE_BUILD_TYPE=Debug \
       -DCMAKE_PREFIX_PATH="$OPENXR_DIR"
--- a/src/xrt/compositor/gl/comp_gl_compositor.cpp
+++ b/src/xrt/compositor/gl/comp_gl_compositor.cpp
@@ -88,6 +88,12 @@ static const IID IID_ID3D11Texture2D_local = {
 #include <OpenGL/OpenGL.h>
 #include <OpenGL/gl3.h>
 #include "comp_gl_window_macos.h"
+#elif defined(XRT_OS_LINUX)
+// Linux desktop: GLAD-provided GL types via the EGL path (mirrors Android).
+// CANDIDATE PATCH (#706 Linux validation) — the original block had no Linux
+// branch, so GL types were undefined and this TU failed to compile.
+#include "ogl/ogl_api.h"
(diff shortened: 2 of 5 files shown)


dfattal (approved):
## Approve — merge

Reviewed with full context (these are the Linux port files). All 5 fixes are correct, minimal, and well-commented, and CI is green. The headline: this closes out the **on-hardware validation** — hosted + handle (`XR_EXT_xlib_window_binding`) sessions bring up the native VK compositor and render the stereo cube, `--service` builds. That's **Phase 1b/2b/3b validated** on real hardware. 🎉

**Per file:**
- `comp_vulkan.c`, both `cube_*_vk_linux/main.cpp` — `#ifdef`-guarding the MoltenVK portability-enumeration macro/flag is exactly right (it's a portability-only opt-in; older Vulkan-Headers like Ubuntu v204 don't define it, macOS/MoltenVK still does). CI missed it only because CI's headers are newer.
- `build_linux.sh` — nested `find` fallback for the app that moved to `test_apps/legacy/` (the #692/#704 regroup) is correct + fails clearly.
- `comp_gl_compositor.cpp` — Linux/EGL branch is correct and **safe** (the Android `#elif` precedes it, so Android isn't captured).

**Two non-blocking follow-ups (filed separately, not holding this merge):**
1. The `comp_gl` Linux branch is safe *by ordering* only — Android also defines `XRT_OS_LINUX`, so `XRT_OS_LINUX_DESKTOP` / `&& !defined(XRT_OS_ANDROID)` would be reorder-proof (matches our `vk_native` convention). It's also compile-only (no EGL context creation wired) — fine, since no Linux app uses GL (all Vulkan).
2. Runtime `build-linux.yml` installs no GL dev libs → `comp_gl` isn't compiled on Linux in CI → this fix isn't regression-guarded. Follow-up: add GL dev libs to CI, or gate `comp_gl` off on Linux.

Thanks for the thorough hardware pass + the clear `CANDIDATE PATCH` annotations.

Issue #709, opened 2026-07-06
comp_gl on Linux: reorder-proof the branch + close the CI-coverage gap (or gate it off)


dfattal:
Follow-up from #708 (Linux hardware validation). Two related, non-urgent gaps around the OpenGL compositor on Linux — the runtime is Vulkan-only on Linux, so this is hygiene, not a functional blocker.

### 1. The Linux branch is safe *by ordering* only
`comp_gl_compositor.cpp` gained `#elif defined(XRT_OS_LINUX)` branches (#708). They're correct **only because** the `#elif defined(XRT_OS_ANDROID)` branch precedes them (Android also defines `XRT_OS_LINUX`). A reorder would silently capture Android into the Linux/EGL branch. Fix: gate on `XRT_OS_LINUX_DESKTOP` / `defined(XRT_OS_LINUX) && !defined(XRT_OS_ANDROID)`, matching the `vk_native` convention.

Also note the branch is **compile-only** — it declares `egl_display/context/surface` but wires no EGL context creation for X11. That's fine today (no Linux app uses GL; all demos + cubes are Vulkan), but it's a dead half-branch.

### 2. CI doesn't compile it
Runtime `build-linux.yml` installs no GL/EGL dev libs → `XRT_HAVE_OPENGL` is OFF → `comp_gl` is **not compiled on Linux in CI**. #708's fix built on real hardware (GL libs present) but a future regression to that branch wouldn't be caught.

### Options (pick one)
- **(a) Cover it:** add GL/EGL dev libs to `build-linux.yml` so `comp_gl`-on-Linux is CI-compiled, and reorder-proof the guard (#1).
- **(b) Gate it off:** since no Linux app uses GL and there's no working GL-on-X11 window backend, don't build `comp_gl` on Linux at all (cleanest — removes the dead branch). Verify no other Linux consumer needs `comp_gl`/`aux_ogl` first.

Recommendation: **(b)** unless a Linux GL app is on the roadmap.

dfattal:
[...]

PR #712, opened 2026-07-06
feat(linux,#660): opt-in software-GL (llvmpipe) present path — XRT_FEATURE_GL_LINUX_SOFTWARE

leaiss:
## TL;DR — an *opt-in software-GL* present path for Linux, not a competitor to VK/XCB

Since this branch was cut, `main` shipped the full Linux port on **Vulkan + XCB** — including the handle class via David's `XR_EXT_xlib_window_binding` (`19c4e014f`, Phase 3a), hardware-validated on a GPU box, plus the deliberate #709 decision to gate `comp_gl` **off** on Linux. **That is and should remain the production Linux path.**

This PR adds a **complementary, default-OFF** capability the VK/XCB path can't provide: an **on-screen present path that runs on software GL (llvmpipe) with no GPU** — for no-GPU CI / headless dev / WSL. Filing it so @dfattal can decide whether that niche is worth carrying.

## Why a GL path can do this and VK/XCB can't

The VK native compositor needs `VK_KHR_external_{memory,semaphore,fence}_fd` to share the app's swapchain image across the app/compositor Vulkan devices — which **software Vulkan (llvmpipe) does not expose** (`vkCreateDevice` → `VK_ERROR_EXTENSION_NOT_PRESENT`). So the production path is correctly GPU-only.

This GL compositor instead **shares the app's GLX context** and samples the app's swapchain textures directly — **no external-memory FD interop** — so it runs end-to-end on llvmpipe. Validated on WSL (Ubuntu, llvmpipe, WSLg): a 4-view anaglyph cube presents on-screen with zero GPU, in both the hosted and handle classes.

## The switch: `XRT_FEATURE_GL_LINUX_SOFTWARE` (default OFF)

`76acc33` makes the whole GL-on-Linux path opt-in behind one CMake option, so the **default Linux build is unchanged** (VK/XCB; no `comp_gl`, no XLIB):

| | OFF (default) | ON (`-D…=ON` / `build_linux.sh --gl`) |
|---|---|---|
| `comp_gl` on Linux | not built (mirrors #709 allowlist) | built |
| `XR_USE_PLATFORM_XLIB` | not armed | armed (Xlib GL binding + `XR_EXT_xlib_window_binding`) |
| GL-native compositor arm | compiled out | `XRT_HAVE_GL_NATIVE_COMPOSITOR` |
| GL test apps | skipped | built |

Verified both ways on WSL: OFF → no `libcomp_gl.a`, no `XR_USE_PLATFORM_XLIB`, selftest PASSED; ON → both present, selftest PASSED.

## Commits

1. `a8bd608` — native GL compositor + GLX present (hosted / self-created window)
2. `cb9d6c4` — qwerty input for the GL compositor
3. `0d1e4c8` — handle class: app-owned X11 window + `cube_handle_gl_linux`
4. `76acc33` — `XRT_FEATURE_GL_LINUX_SOFTWARE` opt-in gate (this makes it coexist with VK/XCB)

## Honest status — one reconciliation item left before merge

- [x] **Opt-in gate** so it doesn't touch the default VK/XCB build — done (`76acc33`).
- [ ] **Adopt main's `XR_EXT_xlib_window_binding`, drop mine.** I independently reinvented the extension with a **conflicting** struct-type (`1000999006` vs main's `1000999200`) and a different struct (`window`-only vs David's `Display*`+`window`). The GL path should consume **David's on-main header** unchanged; my duplicate should be deleted and the GL binding read `xDisplay`+`window` from his struct.
- [ ] **Rebase onto `main`** (currently behind, needs the extension reconciliation above first).

I've held the rebase deliberately — it's only worth doing if we want the software-GL niche. @dfattal: is a no-GPU software-GL present path useful for CI/headless, or should we close this and rely on VK/XCB + a GPU runner?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,6 +208,10 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 	if(PKGCONFIG_FOUND)
 		pkg_check_modules(XCB xcb xcb-randr x11-xcb)
 	endif()
+	# X11 (Xlib) for the native GL compositor's GLX present path (Phase 1b-GL).
+	# GLX itself comes from find_package(OpenGL) above (OpenGL::GLX /
+	# OpenGL_GLX_FOUND). Not REQUIRED — a headless/GL-less box still builds.
+	find_package(X11)
 endif()
 
 # This one is named differently because that's what CTest uses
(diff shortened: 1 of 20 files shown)


dfattal:
Thanks for building this out, Suki — the work is clean and the opt-in gating was done right (default Linux build stays on Vulkan/XCB, verified green).

Closing as a scope/maintenance call, not a quality one. As we discussed: this software-GL (llvmpipe) path is a temporary dev aid for running on WSL while you wait for the target Linux box, and it isn't something we'd carry for real production. Merging it into main would permanently re-enable comp_gl on Linux (which #709 deliberately turned off) and add a 5th compositor backend to maintain — not worth it for a stopgap that the incoming hardware makes moot in a week or two.

The branch (feat/linux-gl-handle-app) stays on GitHub, so you can keep building from it locally on WSL in the meantime. Once the device (or a discrete-GPU-disabled Linux box on the integrated GPU) is in hand, we're back on the real VK/XCB + GPU path, which is where the product runs anyway.

If we ever decide we want automated on-screen render smoke-tests in CI without a GPU runner, this is the branch to revisit.

Issue #730, opened 2026-07-09
Linux vk_native: get_window_metrics returns false for handle apps (app-provided xlib window) — falls back display-scoped


dfattal:
## Problem

The Linux arm of `comp_vk_native_compositor_get_window_metrics()` bails when the compositor doesn't own the window:

- `comp_vk_native_compositor.c:4025` — `if (c->xcb_window == NULL) return false;`
- `comp_vk_native_compositor.c:3459` — handle-class apps (window supplied via `XR_EXT_xlib_window_binding`) set `c->xcb_window = NULL` and carry only `c->xcb_handle` (connection + window).

So for the **dominant app class** on Linux, window metrics fall back display-scoped: window-scoped canvas/Kooima doesn't track the app window's live position/size, and the future window-scoped weave phase (contract §8 R-W7, once the SDK grows the phase-origin knob) has no position source.

## Why this is mechanical

All the pieces already exist for the app-provided handle:

- The **resize path already polls the app window** through `c->xcb_handle`: `comp_vk_native_compositor.c:1053-1056` calls `comp_vk_native_window_xcb_query_geometry(&c->xcb_handle, …)` (same again at `:3546-3550`).
- The **position read** (`xcb_translate_coordinates` → root coords) in `comp_vk_native_window_xcb_get_screen_position()` (`comp_vk_native_window_xcb.c:410-434`) takes a connection + window + root — nothing about it requires the runtime-owned wrapper. Under XWayland (the 26.04 GNOME target) mutter keeps X11 geometry truthful, so the read returns real global coordinates (verified on George's box in the #715 runs).

## Fix

In the `XRT_OS_LINUX_DESKTOP` arm of `get_window_metrics`: when `c->xcb_window == NULL` but `!c->owns_window && c->xcb_handle.connection != NULL`, source window size from `comp_vk_native_window_xcb_query_geometry(&c->xcb_handle, …)` and position from a handle-based variant of the `translate_coordinates` helper (add `comp_vk_native_window_xcb_query_screen_position(const struct comp_vk_native_xcb_handle *, …)` beside the existing query_geometry). Display info path (DP pixel-info → sys_info fallback) is shared and unchanged. Mirrors what the macOS arm already does for the app's external NSView (#524).

## Validation

[...]

Issue #103, opened 2026-03-27
Refactor D3D11 service compositor to match in-process path


dfattal:
## Problem

The D3D11 service compositor (`comp_d3d11_service.cpp`) has accumulated divergent code that reimplements functionality already available in the in-process D3D11 compositor or canonical math libraries. This makes maintenance harder and introduces subtle behavioral differences.

### 1. Canvas params always (0,0,0,0) — **Critical**
Service passes zeros to `process_atlas()` while in-process passes actual canvas bounds. Breaks weaving phase alignment when canvas differs from display. Fix: pass canvas bounds from session/client state.

### 2. Replace IPC hand-rolled Kooima with canonical functions — **Medium**
`ipc_compute_kooima_fov()` + inline m2v scaling in `ipc_server_handler.c` duplicates `display3d_compute_views()` / `camera3d_compute_views()`. Different fallback behavior for edge cases. Fix: call canonical functions from `m_display3d_view.h`.

### 3. Unify crop-blit with in-process path — **Medium**
Service reimplements atlas crop logic inline (~90 lines) instead of reusing `d3d11_crop_atlas_for_dp()` pattern from the in-process compositor. Fix: extract shared crop function or replicate the pattern.

### 4. Generalize IPC to support N views (multiview) — **Medium**
`ipc_try_get_sr_view_poses()` bails on `view_count != 2`. In-process supports up to 8 views. Fix: use eye position arrays and loop over `eye_count`, use `camera3d_compute_views()` / `display3d_compute_views()` (which already support N views).

### 5. Fix HUD eye count filtering — **Low**
Service uses raw `eye_pos->count` without filtering to active rendering mode's view_count, unlike in-process which filters. Fix: add mode filtering.

### 6. Unify zero-copy SRV creation — **Low**
Service manually creates SRVs instead of using swapchain helpers. Fix: align with in-process helper pattern where applicable.

### 7. Sync view dims in sync_tile_layout — **Low** (partially done)
`sync_tile_layout()` now recomputes `view_width/height` from `display_width/tile_columns`, but this was a targeted fix. Consider whether a renderer abstraction (like in-process) would be cleaner long-term.

## Context

Items 1-4 are functional improvements. Items 5-7 are code quality / consistency. The recent mono blit removal (routing both 2D and 3D through the DP) was a first step in this direction.

## Affected Files

- `src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp`
[...]

Issue #101, opened 2026-03-26
Leia SR D3D12 weaver: DPI-unaware HWND causes interlacing mismatch with DPI-aware apps


dfattal:
## Problem

When a DPI-aware app (e.g., Unity 6 with D3D12) passes a canvas rect in physical pixels via `xrSetSharedTextureOutputRectEXT`, the Leia SR D3D12 weaver produces incorrect interlacing. The weaving pattern doesn't match the actual pixel grid.

## Root Cause

The Leia SR weaver internally queries the HWND to determine screen-space positioning for interlacing alignment. On a DPI-aware app with 250% scaling (4K display):

- **App sends:** canvas = 2200×950 (physical pixels) via `xrSetSharedTextureOutputRectEXT`
- **HWND reports:** `GetClientRect()` = 880×380 (logical pixels, because Unity is DPI-aware)
- **Weaver assumes:** the HWND dimensions match the canvas dimensions

This mismatch means the weaver computes the interlacing pattern for an 880×380 pixel area but the actual output is 2200×950 physical pixels. The lens pitch calculations are wrong by a factor of 2.5×.

## Context

The D3D12 `_texture` reference test app (`cube_texture_d3d12_win`) works correctly because it's NOT DPI-aware — its HWND dimensions are physical pixels, matching the canvas rect.

## Expected Fix

The Leia SR D3D12 weaver (or the `xrt_display_processor_d3d12` interface) should:
1. Call `GetDpiForWindow(hwnd)` to determine the DPI scale factor
2. Scale `GetClientRect()` results to physical pixels before comparing with the canvas rect
3. OR use the canvas rect from `process_atlas()` directly for sizing, trusting the app's values

This is a Leia SR weaver/display processor issue, not an OpenXR compositor issue — the compositor correctly passes the canvas rect and HWND through to the display processor.

## Diagnostic Data

```
Canvas (from xrSetSharedTextureOutputRectEXT): 2200×950 (physical px)
HWND GetClientRect: 880×380 (logical px)
DPI scale: 2.50 (GetDpiForSystem = 240)
Display: 3840×2160, Leia SR 3D display
```

## Related

- Unity plugin issue: DisplayXR/displayxr-unity#41
- The plugin correctly sends physical pixel dimensions — this is a runtime/weaver-side DPI handling issue

dfattal:
[...]

PR #758, opened 2026-07-15
feat(linux,#757): X11 ARGB transparent surface, windowed weaving, Wayland backend + compose-under seam

dfattal:
Runtime half of the Linux avatar-overlay transparency feature (#757). **Draft — WIP**; opened for Linux tri-LTS CI compile-validation while the rest of the feature lands.

**In this PR (WS3 + WS5):**
- `XR_DXR_xlib_window_binding`: `transparentBackgroundEnabled` (SPEC_VERSION 1→2), parsed in `oxr_session.c`.
- `comp_vk_native_window_xcb.c`: 32-bit ARGB visual + colormap when transparency is requested (opaque fallback + WARN otherwise); frees colormap on destroy. The target's PRE_MULTIPLIED/INHERIT compositeAlpha selection (already non-macOS-gated) engages once the surface advertises it.

Requires a compositing WM (GNOME/Mutter, KWin, picom).

**Not yet in this PR (tracked on #757):** WS3b Wayland surface backend; extension header sync to displayxr-extensions (on merge).

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -104,7 +104,7 @@ jobs:
             build-essential cmake ninja-build pkg-config \
             libvulkan-dev vulkan-validationlayers glslang-tools \
             libeigen3-dev libcjson-dev libudev-dev \
-            libxcb1-dev libxcb-randr0-dev libx11-dev libx11-xcb-dev \
+            libxcb1-dev libxcb-randr0-dev libx11-dev libx11-xcb-dev libwayland-dev \
             libgl1-mesa-dev libegl1-mesa-dev libxrandr-dev libxxf86vm-dev libxcb-glx0-dev
 
       - name: Build + headless selftest + test app
@@ -139,7 +139,7 @@ jobs:
             build-essential cmake ninja-build pkg-config git \
             libvulkan-dev vulkan-validationlayers glslang-tools \
             libeigen3-dev libcjson-dev libudev-dev \
-            libxcb1-dev libxcb-randr0-dev libx11-dev libx11-xcb-dev \
+            libxcb1-dev libxcb-randr0-dev libx11-dev libx11-xcb-dev libwayland-dev \
             libgl1-mesa-dev libegl1-mesa-dev libxrandr-dev libxxf86vm-dev libxcb-glx0-dev
 
       - uses: actions/checkout@v5
@@ -202,7 +202,7 @@ jobs:
             build-essential cmake ninja-build pkg-config \
             libvulkan-dev vulkan-validationlayers glslang-tools \
             libeigen3-dev libcjson-dev libudev-dev \
-            libxcb1-dev libxcb-randr0-dev libx11-dev libx11-xcb-dev \
+            libxcb1-dev libxcb-randr0-dev libx11-dev libx11-xcb-dev libwayland-dev \
             libgl1-mesa-dev libegl1-mesa-dev libxrandr-dev libxxf86vm-dev libxcb-glx0-dev
 
       - name: Build service + IPC-client runtime (headless selftest)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,6 +441,15 @@ if(XRT_HAVE_SDL2)
 endif()
 
 # Vulkan flags for the shared Vulkan code.
+#
+# These MUST be set here (globally, flowing into xrt_config_vulkan.h) and never
+# as per-target compile definitions: struct vk_bundle (vk_helpers.h) has
+# members conditionally compiled under VK_USE_PLATFORM_*, so a target-local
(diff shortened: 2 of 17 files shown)


dfattal:
Triage (2026-07-25): this branch is **CONFLICTING** (56 behind main) so it can't merge as-is. Note for scope: the **basic X11-ARGB transparent overlay now works on current main** — the avatar demo runs with `AVATAR_TRANSPARENT=1`, creating a 32-bit ARGB app window that composites over the desktop, no crash (the earlier `vkAllocateCommandBuffers` SIGSEGV was a stale-build artifact of this branch: Debug + pre-#785/#730). So this PR's remaining unique value is the **compose-under-bg** model (runtime-side desktop capture + composite in the weave), not the basic transparent surface. Recommend: rebase onto main, drop whatever's now redundant, keep the compose-under-bg work. Left open as active WIP.

dfattal:
**Rebased onto main + trimmed (WIP).** Dropped the now-merged/superseded pieces, kept only the genuinely-unique transparency/Wayland/windowed-weave value. 4087→478 additions, 26→17 files.

**Kept (4 commits):** X11 ARGB transparent surface · WS3b Wayland backend + `XR_DXR_wayland_surface_binding` · present-origin windowed weaving (`set_present_origin`) · XID→SR-weaver view-swap fix.

**Dropped:** cube_zones test (main has all 6 files via #788) · window-metrics commit (main has `query_screen_position` via #730/#787) · stale "zones SW/HW SDK-blocked" doc + comment (contradicts plugin #103, which wired `get_local_zone_caps` and HW zones are ACTIVE) · cli panel-refresh + diag-throttle (#778 scope creep).

Still WIP: compose-under-bg doesn't render transparently on Linux yet. Coupled plugin side = #98 (also trimmed). Verifying the transparency crash resolution on the Suzhou box now that both sit on main (inherits #785 Release + #101 geometry + #730).

dfattal:
**Session-create SIGSEGV root-caused and FIXED** (`cd37328`) — the branch now runs end-to-end on the Suzhou Intel Arc/Odyssey box, **including compose-under transparency with the #98 plugin.**

## The bug
The WS3b commit set `VK_USE_PLATFORM_WAYLAND_KHR` as a compile definition **PRIVATE to `comp_vk_native`**. `struct vk_bundle` (vk_helpers.h:291) conditionally compiles `PFN_vkCreateWaylandSurfaceKHR` under that macro — so `comp_vk_native` saw a bundle one pointer **larger** than `aux_vk`, which fills it (`vk_init_from_given`). Every member after the conditional one was read 8 bytes off: the compositor's `vkCreateCommandPool` slot actually held `vkAllocateCommandBuffers`, invoked with CreateCommandPool's args → the ICD reinterpreted the pool-create-info as an alloc-info, dereferenced a garbage pool handle → **SIGSEGV inside `vkAllocateCommandBuffers` at the `vkCreateCommandPool` call site, on every vk_native session-create** (opaque and transparent alike) on any box where pkg-config finds `wayland-client`. CI never caught it because the Linux jobs are build-green only.

Isolated by a 4-commit on-box bisect (ARGB=good, **wayland=first bad**) after a config matrix showed: tx-runtime crashes with either plugin/mode; main runtime works with either plugin, even transparent. Yesterday's "stale-build crash" was this same bug — the untrimmed branch carried the same commit.

## The fix
Route the macro through the repo's existing global mechanism, like XCB/WIN32/METAL: top-level `set(VK_USE_PLATFORM_WAYLAND_KHR TRUE)` under `XRT_HAVE_WAYLAND` → flows into `xrt_config_vulkan.h` (the `#cmakedefine` slot already existed); drop both target-local defines. Bonus correctness: `XRT_HAVE_WAYLAND` now reaches the oxr target via `xrt_config_have.h`, so `XR_DXR_wayland_surface_binding` is *actually advertised* (the private define never reached `oxr_extension_support.h`'s gate).

**Rule worth keeping:** never set a `VK_USE_PLATFORM_*` macro per-target — `vk_bundle` members are conditionally compiled under them, and the bundle also crosses the runtime→plugin factory boundary raw, so the same consistency requirement applies to plugin builds (a plugin built against pre-fix headers exhibited the mirror-image skew — crash at first `vk->` call in `process_atlas` — until reconfigured).

## Verified on hardware (Intel Arc / Mesa 25.2.8 / Odyssey)
- opaque avatar: 2/2 clean (was 0/2)
- transparent + main plugin: 2/2 clean
- **transparent + #98 compose-under plugin: runs stably** — `transparency = compose-under-bg`, post-weave alpha-gate initialized, zones alpha-over + Local2D bubble active, real srSDK weave, XID→weaver
- WS1 live-desktop capture gracefully degrades to the 2D-under backdrop when the plugin is built without `libpipewire-0.3-dev`/`libdbus-1-dev` (as on the box) — install those to light it up

Issue #625, opened 2026-06-20
Step 0: DXR weave RPC (bindWindow + weave over IPC, browser-free)


dfattal:
## Summary

Implement **Step 0** of the inline-3D-in-the-browser roadmap (`docs/roadmap/webxr-support.md` §2.4): a window-bound **DXR weave RPC** in `displayxr-service`, with **zero browser / no CEF**. This is the load-bearing new runtime capability — a synchronous weave *service* distinct from the steady swapchain frame loop — and the first milestone the later CEF host (Step A) and Chromium patch (Step B) build on.

```
bindWindow(handle)
weave(shared_handle, window_relative_rect, eyes) -> weaved_shared_handle
```

The runtime imports the caller's **pre-weave** SBS texture, runs the **DP** weave (`process_atlas`) at the phase derived from window-position + rect, and exports the weaved handle for the caller to composite/present itself (present-ownership). The browser/caller **never weaves** (ADR-007 / ADR-019). Reuses the existing **texture-class + window-binding + display-zones + DP weave** machinery — no new weave/interlace code.

## Phase 0 — gap analysis (audited, file:line)

- **Texture-class present-ownership / weave-into-caller-shared-texture is in-process only.** `oxr_xrSetSharedTextureOutputRectEXT` / `…Surround2DEXT` dispatch only to native-compositor branches (`sess->xcn`) and return `XR_ERROR_FEATURE_UNSUPPORTED` over IPC — `oxr_api_session.c:1643-1678`.
- **The one weaved-pixels-back-to-a-present-owning-client path** is the transparent-output route (#551): server weaves into a **server-allocated** single full-frame shared texture + fence, hands back handle + the app's HWND, client DComp-presents — `comp_d3d11_service.cpp:11261`, `ipc_server_handler.c:5286`, `ipc_client_compositor.c:402`, `proto.json:692`. No output-rect, no caller input texture, no zones, D3D11-only.
- **`process_atlas` carries no eye/IPD fields** (`xrt_display_processor.h:124-141`); the DP weaves from eyes it reads internally from the vendor tracker. Eyes flow DP→app, never app→DP — so caller-supplied eyes driving the weave is the genuinely new "external eyes" plumbing.
[...]

PR #775, opened 2026-07-20
feat(weave): N-view worst-case atlas input (spec v6) — #774

dfattal:
Closes #774.

## What

Teaches the `XR_DXR_weave` batch path the atlas layout every other DisplayXR app already uses, so a present-owner (the browser) becomes an ordinary N-view client instead of the one special case.

A chained `XrWeaveSubmitLayoutDXR` (type `1004999194`) declares a **worst-case-sized N-view atlas** — sized from the *display* across all rendering modes (ADR-010), filled per-frame with tiles packed **contiguously from the top-left** at the active mode's content view size (`window × mode.viewScaleX/Y`).

## Why

The batch path hardcoded not just `view_count = 2` but **`view_scale = (0.5, 1.0)`**. It allocated a `2·win_w × win_h` scratch and ran **2 stretch-blits per rect** to upsample each squeezed half back to full width, then declared a 1.0-scale tile carrying 0.5-scale pixels — which the DP resampled straight back down. Waste on every frame, proportional to rect count, and structurally unable to express any vendor mode with a different grid or per-view scale.

## How

On the v6 path the runtime does **less**: no SBS scratch, no per-rect unpack blits. The packed region goes to the DP as-is when it exactly fills the active mode's atlas (zero-copy per ADR-030), otherwise **one box copy** crops it — contiguous packing makes the packed region a single top-left rectangle, unlike the per-tile gather the `xrEndFrame` path needs.

`process_atlas()` has always taken an arbitrary N-tile atlas, so there is **no DP ABI bump**. Callers that omit the chain keep v3/v4/v5 behaviour byte-for-byte.

## Two subtleties worth reviewing

- **The worst-case atlas is not bounded by the display size.** A mode with `viewScaleX = 1.0, tileColumns = 2` yields `2W × H`. Confirmed on hardware: Leia reports `SR recommended view texture: 1920x1080 per eye` on a 1536×864 panel.
- **Tile stride is the CONTENT size, not `atlasWidth / tileColumns`.** The latter is the shell/multi-compositor invariant (`multiview-tiling.md`), for atlases the *runtime* writes. The app rule is `renderW = window * scaleX` at `tileX * renderW` (INV-4.3/4.4); `swapchain.width / tileColumns` is only a clamp.

## Validation

- Runtime build clean; `displayxr-cli selftest` **PASS** against the live Leia SR plug-in (ABI v4 still loads).
- `weave_rpc_probe_d3d11_win` gated on `DXR_WEAVE_V6=1`, with the atlas deliberately allocated 2× the packed region so the **crop** branch is under test rather than zero-copy:
  ```
  #625 weave v6: crop staging 1280x720 (atlas 2560x720, 2 view(s) 640x720 grid 2x1)
  #625 weave v6: views=2 grid=2x1 content=640x720 atlas=2560x720 win=1280x720 zc=0
  ```
- Element geometry lands at the correct window position; the v4 overlay atlas still composites crisply over the weave.
- **Live Leia 3D eyeball confirmed good.**

Caveat on the v5-vs-v6 image diff: it is *not* a pixel-equivalence proof, because live eye tracking moves the parallax squares between captures so the two runs are not frame-matched. It only establishes that the difference is confined to the element rect (gaps and overlay bit-identical).

## Not in this PR

Browser-side work (worst-case allocation from `xrEnumerateDisplayRenderingModesDXR`, reallocation on *display* change rather than window resize, and sub-regioning view *v* into tile *v* in `skia_output_surface_impl_on_gpu.cc`) is separate, in `displayxr-browser`. Landing at N = 2 first is intentional: same content budget as today, so it ships as a simplification with no visual change.

https://claude.ai/code/session_01J8tEc8D6wZyaDznvnZmLtt


dfattal:
**Both v6 branches now hardware-eyeballed on live Leia, post-fix.**

Worth calling out explicitly because the original validation in the PR body predated the
zero-copy sync fix (`20ab47a8c`), and that fix touches code shared by both branches — so the
earlier eyeball no longer covered what shipped.

| run | branch | log |
|---|---|---|
| `DXR_WEAVE_V6=1 DXR_WEAVE_V6_EXACT=1` | zero-copy | `views=2 grid=2x1 content=640x720 atlas=1280x720 win=1280x720 zc=1` |
| `DXR_WEAVE_V6=1` | crop | `crop staging 1280x720 (atlas 2560x1440 ...)`, `zc=0` |

Run back to back, non-elevated against the same service. Both show correct glasses-free 3D —
depth separation and head-tracked look-around, on a transparent background, with the v4 2D
overlay flat at screen depth. Reported identical between the two branches.

This closes the coverage gap that hid the sync bug: until `DXR_WEAVE_V6_EXACT=1` existed,
every run took the crop branch and zero-copy had never executed once.

Caveat for review: the runtime internals are now well exercised, but every defect found in
this PR after the initial "validated" claim (zero-copy sync, `SPEC_VERSION`, the v6
compatibility gate) came from the mac box implementing against the spec — not from testing
here. The probe constructs the chain directly and ignores `extensionVersion`, so it
structurally cannot catch contract-level mistakes.

Issue #225, opened 2026-05-15
DP compose-under-bg likely applied per-layer instead of post-composite — desktop bleeds through semi-transparent window-space HUD over opaque content


dfattal:
## Symptom

When a semi-transparent `XrCompositionLayerWindowSpaceEXT` (HUD / tuning panel) is composited over an opaque pixel in the projection layer (e.g. the tiger character in `displayxr-unity-test-transparent`), the user sees the **captured desktop background bleeding through the panel** instead of the opaque content underneath.

The HUD is visually on top of the tiger, but where you'd expect the tiger to show through the panel's transparency, you see the desktop background mixed into the panel color.

Likely related symptom: the tiger does not appear in Windows screen capture (Snipping Tool / Win+Shift+S) — only the desktop is captured, the overlay's content is missing.

## Repro

Using `displayxr-unity@fix/clickthrough-nc-messages` (or any v1.6+ plugin) and `displayxr-unity-test-transparent`:

1. Run the test app on hardware (transparent overlay enabled).
2. Tiger appears over the desktop. ✓
3. Press Shift+Tab to bring up the `TigerTuningHUD` (a `DisplayXRWindowSpaceUI` window-space layer with a semi-transparent panel).
4. Position the panel so it overlaps the opaque body of the tiger.
5. **Expected:** through the panel's semi-transparency, you see the tiger's body (50% panel + 50% tiger).
6. **Actual:** through the panel's semi-transparency, you see the captured desktop (50% panel + 50% desktop), as if the tiger isn't there at those pixels.

## Plugin side ruled out

Verified the plugin is doing the right thing:

- `DisplayXRWindowSpaceUI`'s overlay camera clears the wsui swapchain to `Color.clear` = (0,0,0,0). Confirmed in `Runtime/DisplayXRWindowSpaceUI.cs:169-170`. So the wsui RT has alpha=0 outside the panel, panel-alpha inside.
- The plugin submits the window-space layer with `XR_COMPOSITION_LAYER_BLEND_TEXTURE_SOURCE_ALPHA_BIT | XR_COMPOSITION_LAYER_UNPREMULTIPLIED_ALPHA_BIT` (the UNPREMULTIPLIED flag was added today in DisplayXR/displayxr-unity@7e9ef4d trying to fix this — didn't help, confirming the bug is downstream).
[...]

PR #779, opened 2026-07-21
fix(d3d11-service): honor layered swapchains in zones/Local-2D composite (ADR-032)

dfattal:
## Problem — DisplayXR/displayxr-unity#225

With the workspace-tile provider path (displayxr-unity#223), BiRP/URP/HDRP samples render correctly as shell tiles, but **desktop-avatar renders flat/distorted**. It's the only sample using `XR_DXR_display_zones` (a 3D zone) + a Local2D bubble.

## Root cause — the ADR-032 gap #767 left in the zones path

#767 ported the ADR-032 array-slice fix to the service's **primary projection** blit, but `service_composite_zones_frame` — the `XRT_LAYER_ZONE_3D` + `XRT_LAYER_LOCAL_2D` path — was never updated. It sampled a plain `Texture2D` SRV and **never read `sub.array_index`**, so a **layered (arraySize=2 SPI)** zone source (the render path the Unity provider uses for URP/HDRP) had **both eyes sampled at slice 0 → no disparity → flat**, and the zone rendered outside the primary-projection slot handling → distorted.

Confirmed from three angles:
- **Plugin** forwards all layers via `xrEndFrame` unconditionally (the zone chains onto the primary projection's `.next`; `submit_zone_3d_layer` turns it into `XRT_LAYER_ZONE_3D` with per-view `sub.array_index`). No plugin bug.
- **IPC** carries every typed layer with `array_index`. No drop.
- **Service** `service_composite_zones_frame` had no array-slice branch.

The **native VK avatar demo works** because it submits **tiled (arraySize=1, per-view `imageRect`)** swapchains, which this path already handles. The sole difference is tiled-vs-layered swapchain geometry — not layer topology (both submit zone-on-`proj.next` + separate Local2D, 2 layers).

## Fix (mirror of #767 into the zones composite)

- **zone-3D blit:** `is_layered = sd.ArraySize > 1`, `src_slice = proj.v[v].sub.array_index`; pass `is_array`/`array_slice` into `blit_to_atlas_texture`. The per-image SRV is already a `Texture2DArray` for layered sources (create/import), so the workspace path needs no SRV change.
- **Local-2D blit:** same, for robustness (no-op for the common `arraySize=1` band).
- **`zones_resolve_src_srv`:** build a whole-array SRGB SRV for layered sources so the standalone (non-workspace) sRGB path is array-correct too.
- Tiled (`arraySize=1`) sources stay byte-identical.

## Test
- `comp_d3d11_service` compiles clean (Release).
- **HW verify pending:** version-matched shell + service on the Leia box, desktop-avatar as a workspace tile — expect live stereo + correctly-framed zone + bubble (compare to standalone app-overlay). Single-projection tiles (BiRP/URP/HDRP) must stay unaffected.

Refs ADR-032, #767.

https://claude.ai/code/session_01Umm4stoD3G9SuNeocafhPn

### DIFF
--- a/src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp
+++ b/src/xrt/compositor/d3d11_service/comp_d3d11_service.cpp
@@ -8523,9 +8523,21 @@ zones_resolve_src_srv(struct d3d11_service_system *sys,
 	if (!sys->workspace_mode && is_srgb_format(desc->Format)) {
 		D3D11_SHADER_RESOURCE_VIEW_DESC srv_desc = {};
 		srv_desc.Format = get_srgb_format(desc->Format);
-		srv_desc.ViewDimension = D3D11_SRV_DIMENSION_TEXTURE2D;
-		srv_desc.Texture2D.MipLevels = 1;
-		srv_desc.Texture2D.MostDetailedMip = 0;
+		// ADR-032 (#225): a LAYERED (arraySize>1) source needs a whole-array
+		// SRGB SRV so the array PS can select the slice — matching the plain
+		// per-image SRV, which create/import already builds as a Texture2DArray
+		// for layered sources. Tiled (arraySize==1) sources keep Texture2D.
+		if (desc->ArraySize > 1) {
+			srv_desc.ViewDimension = D3D11_SRV_DIMENSION_TEXTURE2DARRAY;
+			srv_desc.Texture2DArray.MostDetailedMip = 0;
+			srv_desc.Texture2DArray.MipLevels = 1;
+			srv_desc.Texture2DArray.FirstArraySlice = 0;
+			srv_desc.Texture2DArray.ArraySize = desc->ArraySize;
+		} else {
+			srv_desc.ViewDimension = D3D11_SRV_DIMENSION_TEXTURE2D;
+			srv_desc.Texture2D.MipLevels = 1;
+			srv_desc.Texture2D.MostDetailedMip = 0;
+		}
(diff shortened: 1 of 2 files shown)


dfattal:
## HW-verified ✓ (Leia panel, 2026-07-22)

Built this fix pinned to the exact deployed commit (`87ef20a1c` + cherry-pick → `v2.0.8-5-g4ccaf439a`, IPC-struct-identical to the deployed `v2.0.8-4` client, so `IPC_IGNORE_VERSION=1` is safe), swapped it in for the deployed `displayxr-service.exe`, and launched desktop-avatar as a workspace tile.

**Result:** the tiger (3D zone, arr=2 layered) and the bubble (Local2D) now composite with **correct per-eye stereo** — the flat/no-disparity symptom is gone. Plugin log steady: `display-zones ACTIVE maxZones3D=8`, `3d_zone 904x1088`, `local2d 1024x640 (D3D12 bridge)`, `xrEndFrame layers=2 r=0`, no IPC errors.

Two *separate* workspace-tile follow-ups surfaced during the same test (not caused by this change — they exist for the layered path now that it renders): (1) the app authors its zone/bubble rects for its own saved window size and the service canvas (`GetClientRect(app_hwnd)`) stretches that to the tile slot; (2) keyboard input isn't routed to the minimized tile window. Tracking those independently.

dfattal:
Superseded by #780, which bundled this exact #225 zones/Local-2D layered-slice fix (comp_d3d11_service.cpp) together with the xrGetWorkspaceTileSizeDXR query and is now merged to main. Closing to avoid a duplicate merge.

